PR #5, opened 2020-11-03
Use OpBuilderDAG instead of OpBuilder

marbre:
Takes over the upstream changes llvm/llvm-project@89eab30 to EmitC.


marbre:
Manually fixed this commit, since GitHub messed up the commiter email address.

PR #18, opened 2020-12-08
Bump mlir-hlo and llvm

marbre:
Bump MLIR-HLO submodule to tensorflow/mlir-hlo@812221d and set LLVM
to llvm/llvm-project@92160b2.


simon-camp (approved):
LGTM

PR #34, opened 2021-01-22
Fix gcc warnings

marbre:
So far, gcc moaned that e.g. `struct std::array<long unsigned int, 1>` has no
user-provided default constructor and the implicitly-defined
constructor does not initialize
`long unsigned int std::array<long unsigned int, 1>::_M_elems [1]`.

Co-authored-by: David Ronnenberg <david.ronnenberg@iml.fraunhofer.de>


dajuro (approved):
Looking good :+1:

PR #36, opened 2021-01-25
Pass data to unary/binary templated function as const reference

marbre:
Reduces the size for MobileNetv2 from `15778496` to `15769032` byte.


dajuro:
> Reduces the size for MobileNetv2 from `15778496` to `15769032` byte.

Is this the reduction in size of the executable or the memory consumption?

marbre:
> > Reduces the size for MobileNetv2 from `15778496` to `15769032` byte.
> 
> Is this the reduction in size of the executable or the memory consumption?

It just the binary size, we still should benchmark the memory consumption.

dajuro (approved):
Looking good :+1:

PR #39, opened 2021-01-27
Add keras end2end script

simon-camp:
This makes the end2end flow available as a bash script and removes the deprecated conversion script.


scripts/e2e_test_keras.sh
```diff
@@ -0,0 +1,84 @@
...
+  exit 1
+fi
+
+KERAS_MODEL=$1
+EMITC_INCLUDE_DIR=$2
+EMITC_OPT=$3
+EMITC_TRANSLATE=$(dirname $EMITC_OPT)/emitc-translate
+CPP_COMPILER=$4
+SEED=$5
+BATCH_SIZE=2
```
marbre, line 31: Should we make this an input parameter? In internal use, a colleague hit the issue that he expected a batch size of 1. This can be a followup though.

scripts/e2e_test_keras.sh
```diff
@@ -0,0 +1,84 @@
...
+echo "CPP_COMPILER=$CPP_COMPILER"
+echo "SEED=$SEED"
+echo "BATCH_SIZE=$BATCH_SIZE"
+echo "OUTPUT_DIR=$OUTPUT_DIR"
+
+echo "Setting up output directory"
+mkdir -p "$OUTPUT_DIR"
+
+echo "Copying over keras model"
+cp "$KERAS_MODEL" "$OUTPUT_DIR"/model.h5
```
marbre, line 47: I don't see why copying is strictly necessary. At least it is used inconsistently.

scripts/e2e_test_keras.sh
```diff
@@ -0,0 +1,84 @@
...
+
+echo "Converting keras model to saved model format"
+python keras_to_savedmodel.py --batch-size $BATCH_SIZE "$KERAS_MODEL" "$OUTPUT_DIR"/model
```
marbre, line 50: Here the original model is already used.

scripts/e2e_test_keras.sh
```diff
@@ -0,0 +1,84 @@
...
+sed "s/$FUNCTION_NAME/predict/g" "$OUTPUT_DIR"/model_canon.mlir > "$OUTPUT_DIR"/model_fix_name.mlir
+
+echo "Converting mhlo dialect to emitc dialect"
+"$EMITC_OPT" --mhlo-control-flow-to-scf --convert-mhlo-region-ops-to-emitc --convert-mhlo-to-emitc --convert-scf-to-emitc --convert-std-to-emitc --convert-mhlo-const-to-std "$OUTPUT_DIR"/model_fix_name.mlir > "$OUTPUT_DIR"/model_emitc.mlir 
+
+echo "Translating emitc dialect to cpp header"
+"$EMITC_TRANSLATE" --mlir-to-cpp "$OUTPUT_DIR"/model_emitc.mlir > "$OUTPUT_DIR"/model_generated.h
+
+echo "Generating test case"
+python generate_testscases.py --file-format cpp --count 1 --batch-size $BATCH_SIZE --seed $SEED "$OUTPUT_DIR"/model.h5 "$OUTPUT_DIR"
```
marbre, line 78:
```suggestion
python generate_testscases.py --file-format cpp --count 1 --batch-size $BATCH_SIZE --seed $SEED "$KERAS_MODEL" "$OUTPUT_DIR"
```

marbre (changes requested):
One further remark. Sometimes you quote variables using curly braces, sometimes you don't (`"$VAR"` vs. `"${VAR}"`). In one of the next iterations we can try to make this more consistent, but beside the inline comments it lgtm anyway.

scripts/e2e_test_keras.sh
simon-camp: This is a left over from testing. I'll delete it.

scripts/e2e_test_keras.sh
simon-camp: Done.

simon-camp:
> One further remark. Sometimes you quote variables using curly braces, sometimes you don't (`"$VAR"` vs. `"${VAR}"`). In one of the next iterations we can try to make this more consistent, but beside the inline comments it lgtm anyway.

That was shell paramter extension, to chop of the first character. But you are right that this was not needed :)

marbre (approved):

PR #45, opened 2021-01-28
Remove dimension specific typedefs

simon-camp:
These were leftovers from the move to a variadic template for the Tensor type.


marbre:
As discussed, lets drop the dimension in the header-only implementations an in the printer. But lets keep the typedefs as well as using them in the unittests for easier readability.

marbre (changes requested):
As discussed, lets drop the dimension in the header-only implementations an in the printer. But lets keep the typedefs as well as using them in the unittests for easier readability.

simon-camp:
The only place i kept this apart from the unittests was in the concat meta function.

marbre (approved):

PR #52, opened 2021-02-03
Generalize e2e pipeline test to work with keras and tensorflow saved models

OliverScherf:
The existing test pipeline already worked with both keras and tensorflow saved models, hence I generalized the naming and the content of the scripts. Since I renamed the script, I also updated the github build-and-test.yaml. 

Furthermore the test case generator can now handle int32 as output type of operations.

### DIFF
--- a/scripts/generate_testscases.py
+++ b/scripts/generate_testscases.py
@@ -62,6 +62,8 @@ def save_examples(path: str, examples, file_format: str):
         dtype = array.dtype
         if dtype == np.float32:
             return "float"
+        elif dtype == np.int32:
+            return "int32_t"
         else:
             raise ValueError(
                 f"C type conversion for dtype '{dtype}' not implemented.'")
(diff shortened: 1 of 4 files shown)


scripts/model_to_savedmodel_with_predict_function.py
```diff
@@ -41,24 +41,32 @@ def translate(model_path: str, output_path: str, batch_size: int):
...
 
     tf.saved_model.save(module, output_path)
 
 
 def main():
     parser = argparse.ArgumentParser(
-        description="Translate keras model to saved model format")
+        description=
+        "Translates a model to saved model format with a predict function for further compiling." \
+        "Both keras and saved model format as input is supported."
```
marbre, line 56:
```suggestion
       Keras and saved model format are supported as input.
```

marbre (changes requested):
Minor nit.

scripts/generate_testscases.py
```diff
@@ -120,15 +122,25 @@ def function_name():
                 )
 
             for i, curr_output in enumerate(outputs):
+                if not isinstance(curr_output, np.ndarray):
+                    if isinstance(curr_output, int):
```
simon-camp, line 126: We should check if tf.int_64 is also represented as Python int here. It's worth adding an e2e test for both cases in a follow up.

simon-camp (approved):
LGTM. I'm not sure about the code reformatting though. We may want to a linter configuration file.

marbre (approved):

PR #57, opened 2021-02-18
Introduce core ops header and move mhlo::abs to it

marbre:
This introduces are core ops header in which common ops are implemented.
Those common ops can be called/used from different conversion passes.


dajuro (approved):
LGTM

PR #59, opened 2021-02-18
Add TOSA add and exp operations

dajuro:
Adds `add` and `exp` operations to `emitc_tosa.h` and corresponding tests. To achieve a shared implementation between mhlo and tosa these ops were moved from `emitc_mhlo.h` to `emitc_core_ops.h`.

### DIFF
--- a/docs/tosa-op-coverage.md
+++ b/docs/tosa-op-coverage.md
@@ -6,3 +6,6 @@ The table below shows the supported TOSA ops.
 | :-------------------- |:------------------:| :------ |
 | **Unary elementwise ops**
 | abs                   | :heavy_check_mark: | |
+| exp                   | :heavy_check_mark: | |
+| **Binary elementwise ops**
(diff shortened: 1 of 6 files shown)


marbre:
Maybe the PR title or at least the PR description should mention that the so far mhlo only ops were moved as these are common core functions/ops.

marbre (changes requested):
Please add the now supported ops to `docs/tosa-op-coverage.md`.

dajuro:
I fixed the tests and added the ops to `docs/tosa-op-coverage.md`. But I completely screwed up to do this via the github "requested changes"-workflow...

marbre (approved):
The "requested" changes just means that there is something left to do from the reviewer's perspective. I didn't proposed/made any suggestions. Thus you can fix this with addition commits or change the existing commits (choosing the way you prefer) and request a re-review :)

PR #64, opened 2021-02-26
Add conversions for tosa.reciprocal and tosa.rsqrt.

dajuro:
This adds conversions for `tosa.reciprocal` and `tosa.rsqrt`. Also adds tests for conversions and functions. `tosa.rsqrt` is splitted into two `emitc::CallOp`: `emitc::sqrt` and `tosa::reciprocal`.


marbre (approved):
Awesome. Really happy to see progress :)

PR #65, opened 2021-02-27
Replace deprecated 'getAttrs'

marbre:
'getAttrs' has been explicitly marked deprecated. This patch refactors
to use Operation::getAttrs().


simon-camp (approved):
LGTM.

PR #66, opened 2021-03-01
Add conversions for tosa.fully_connected and tosa.mul

OliverScherf:
* Adds conversions for tosa.fully_connected and tosa.mul
* Adds support for broadcasting which is used implicitly in TOSA's binary ops such as tosa.mul and tosa.add


include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,63 @@ inline Src add(Src x, Src y) {
...
+}
+
+/// Functions for other TOSA ops.
+// BroadcastInDimOp
+template <typename Dest, typename Src>
+inline Dest
+broadcast_in_dim(Src operand,
+                 Tensor<int64_t, Src::rank()> broadcast_dimensions) {
+  return emitc::broadcast_in_dim<Dest>(operand, broadcast_dimensions);
+}
```
marbre, line 66: There is no `broadcast_in_dim` in TOSA. Thus, we shouldn't define one in our header. Just directly convert to an emitc call that is calling the core func.

include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,63 @@ inline Src add(Src x, Src y) {
...
+
+
+
+// FullyConnectedOp
```
marbre, line 70: @simon-camp Please review this one.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,171 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+SmallVector<Value, 2>
+create_broadcast_op_if_needed(SrcOp &srcOp, ArrayRef<Value> operands,
+                              ConversionPatternRewriter &rewriter) {
+  // TOSA allows implicit broadcasting, so we need to insert broadcast_in_dim
+  // Ops if necessary
+  // clang-format off
+  // tosa.add tensor<8xf32>)     -> tensor<1x4x4x8xf32>     Broadcast dims = (3)
+  // tosa.add tensor<4x8xf32>)   -> tensor<1x4x4x8xf32>     Broadcast dims = (2, 3) 
+  // tosa.add tensor<4x4x8xf32>) -> tensor<1x4x4x8xf32>     Broadcast dims = (1, 2, 3)
+  // clang-format on
```
marbre, line 121: Why a clang-format off?

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,171 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
   }
 };
 
+template <typename SrcOp>
+SmallVector<Value, 2>
+create_broadcast_op_if_needed(SrcOp &srcOp, ArrayRef<Value> operands,
+                              ConversionPatternRewriter &rewriter) {
+  // TOSA allows implicit broadcasting, so we need to insert broadcast_in_dim
+  // Ops if necessary
```
marbre, line 116:
```suggestion
  // ops if necessary
```

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,171 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+  // clang-format off
+  // tosa.add tensor<8xf32>)     -> tensor<1x4x4x8xf32>     Broadcast dims = (3)
+  // tosa.add tensor<4x8xf32>)   -> tensor<1x4x4x8xf32>     Broadcast dims = (2, 3) 
+  // tosa.add tensor<4x4x8xf32>) -> tensor<1x4x4x8xf32>     Broadcast dims = (1, 2, 3)
+  // clang-format on
+
+  StringRef broadcastFuncName = "emitc::broadcast_in_dim";
```
marbre, line 123: Here you're actually calling the core op. So as mentioned before, no need to add it to the tosa header.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,171 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+  StringAttr broadcastCallee = rewriter.getStringAttr(broadcastFuncName);
+
+  Value output = srcOp.getResult();
+  auto opOutputShape = output.getType().cast<RankedTensorType>().getShape();
+  auto opOutputRank = output.getType().cast<RankedTensorType>().getRank();
+  SmallVector<Value, 2> broadcastedOperands;
+  broadcastedOperands.push_back(operands[0]);
+  broadcastedOperands.push_back(operands[1]);
+  for (size_t i = 0; i < operands.size(); i++) {
+    auto &operand = operands[i];
```
marbre, line 133: Why not iterating over the operands?

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -122,6 +282,14 @@ void populateTosaToEmitcPatterns(MLIRContext *ctx,
 
   // Binary elementwise ops
   patterns.insert<CallOpConversion<tosa::AddOp>>(ctx, "tosa::add");
+  patterns.insert<CallOpConversion<tosa::MulOp>>(ctx, "tosa::mul");
+
+  patterns.insert<CallOpBroadcastableConversion<tosa::AddOp>>(ctx, "tosa::add");
+  patterns.insert<MulOpConversion>(ctx, "tosa::mul");
```
marbre, line 288: Adding `patterns.insert<CallOpConversion<tosa::MulOp>>(ctx, "tosa::mul");` *and* `patterns.insert<MulOpConversion>(ctx, "tosa::mul");` seems wrong.

test/Conversion/tosa-to-emitc.mlir
```diff
@@ -30,7 +30,63 @@ func @test_rsqrt(%arg0: tensor<13x21x3xf32>) -> tensor<13x21x3xf32> {
...
+
+// MulOp: Second operand needs to be broadcasted + expanded
+func @test_mul4(%arg0: tensor<13x21x3xf32>, %arg1: tensor<3xf32>) -> tensor<13x21x3xf32> {
+  // CHECK: emitc.call "emitc::broadcast_in_dim"(%arg1) {args = [dense<2> : tensor<1xi64>], template_args = [tensor<13x21x3xf32>]} : (tensor<3xf32>) -> tensor<13x21x3xf32>
+  // CHECK: emitc.call "tosa::mul"(%arg0, %0) {args = [1 : i32], template_args = []} : (tensor<13x21x3xf32>, tensor<13x21x3xf32>) -> tensor<13x21x3xf32>
+  %0 = "tosa.mul"(%arg0, %arg1)  { shift = 1 : i32 } : (tensor<13x21x3xf32>, tensor<3xf32>) -> tensor<13x21x3xf32>
+  return %0 : tensor<13x21x3xf32>
+}
+
+// MulOp: Second operand needs to be broadcasted + expanded with two dimension
```
marbre, line 78:
```suggestions
// MulOp: Second operand needs to be broadcasted + expanded by two dimensions
```
By two dimensions or by a second? Haven't looked that closely.

marbre (changes requested):
Some first comments, will look to the conversions more closely later.

marbre:
FYI, normally the C++ unittests belong to the PR and shouldn't be added separately.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: I need the index to map the broadcasted operands to the SrcOp's operands, if a broadcast has to happens.

test/Conversion/tosa-to-emitc.mlir
OliverScherf: I updated the comment in the test to make it more clear:

`// MulOp: Second two dimensional operand needs to be broadcasted + expanded to four dimensions`

A broadcast from a Tensor2D into a Tensor4D is happening here.

include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,52 @@ inline Src add(Src x, Src y) {
...
+  static_assert(output.dim(1) == weights.dim(0),
+                "Output and weights dimensions does not match.");
+  static_assert(weights.dim(0) == bias.dim(0),
+                "Bias and weights dimensions does not match.");
+
+  const size_t N = input.dim(0);
+  const size_t C_IN = input.dim(1);
+  const size_t C_OUT = weights.dim(0);
+
+  for (size_t n = 0; n < N; ++n) {
```
simon-camp, line 85: We use post increment everywhere else I think.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: clang-format would format the comment into more lines and this would reduce clarity. I shortened the comment so it wouldn't add line breaks and removed the clang-format off statement, because there was a typo in the comment anyway.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,169 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+        mlir::ValueRange({broadcastedOperands[0], broadcastedOperands[1]}));
+
+    return success();
+  }
+
+  StringRef funcName;
+  // If set, use the result type of the operation as template parameter
+  bool explicitResultType;
+  // If set, use the operand types as (additional) template parameters
+  bool explicitOperandTypes;
```
simon-camp, line 267: As this conversion operates on a single op, these attributes can be removed (inlined).

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,169 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+                  bool explicitOperandTypes = false)
+      : OpConversionPattern<tosa::MulOp>(ctx), funcName(funcName),
+        explicitResultType(explicitResultType),
+        explicitOperandTypes(explicitOperandTypes) {}
+
+private:
+  LogicalResult
+  matchAndRewrite(tosa::MulOp mulOp, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    StringAttr callee = rewriter.getStringAttr(funcName);
```
simon-camp, line 231: Hardcode `"tosa::mul"`

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,169 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+      Type type = mulOp.getType();
+      templateArgs_.push_back(TypeAttr::get(type));
+    }
+
+    if (explicitOperandTypes) {
+      for (auto &operand : operands) {
+        Type type = operand.getType();
+        templateArgs_.push_back(TypeAttr::get(type));
+      }
+    }
```
simon-camp, line 245: These are always false for the mul operation.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,169 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+
+    SmallVector<Attribute, 1> args_;
+    args_.push_back(mulOp.shiftAttr());
+    ArrayAttr args = rewriter.getArrayAttr(args_);
+
+    ArrayAttr templateArgs = ArrayAttr::get(mulOp.getContext(), templateArgs_);
```
simon-camp, line 251: We don't need explicit template args here. This can be changed to `ArrayAttr templateArgs;`, which cleans up the generated IR.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,169 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+
+    if (explicitOperandTypes) {
+      for (auto &operand : operands) {
+        Type type = operand.getType();
+        templateArgs_.push_back(TypeAttr::get(type));
+      }
+    }
+
+    SmallVector<Attribute, 1> args_;
+    args_.push_back(mulOp.shiftAttr());
```
simon-camp, line 248: When we add additional arguments to the call operation we need to insert an index attribute for all operands first. i.e. something like this (untested).

```c++
args_.push_back(rewriter.getIndexAttr(0));
args_.push_back(rewriter.getIndexAttr(1));
args_.push_back(mulOp.shiftAttr());
```

simon-camp (changes requested):
I've left some comments.

OliverScherf:
Thanks for your feedback @marbre and @simon-camp! I implemented the requested changes.

I also added Unittests for the FullyConnectedOp and MulOp.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
marbre: Please don't mark comments as resolved if you don't address them via commits. Otherwise I have to go through all previous comments...

I hoped one could still use `for (auto operand : operands)` and avoid C style loop, but if not, it's okay.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,149 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+  StringRef broadcastFuncName = "emitc::broadcast_in_dim";
+  StringAttr broadcastCallee = rewriter.getStringAttr(broadcastFuncName);
+
+  Value output = srcOp.getResult();
+  auto opOutputShape = output.getType().cast<RankedTensorType>().getShape();
+  auto opOutputRank = output.getType().cast<RankedTensorType>().getRank();
+  SmallVector<Value, 2> broadcastedOperands;
+  broadcastedOperands.push_back(operands[0]);
+  broadcastedOperands.push_back(operands[1]);
+  for (size_t i = 0; i < operands.size(); i++) {
```
marbre, line 130: You may want to change this to
```suggestion
  for (size_t i = 0; i < operands.size(); ++i) {
```

include/emitc/emitc_core_ops.h
```diff
@@ -71,6 +71,61 @@ inline Src add(Src x, Src y) {
...
+template <typename Src>
+inline Src mul(Src x, Src y) {
+  using ET_Src = typename get_element_type<Src>::type;
+
+  auto f = std::multiplies<ET_Src>{};
+
+  return binary<Src>(x, y, f);
+}
+
+/// Functions for other ops.
```
marbre, line 84:
```suggestion
/// Other ops
```

include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,72 @@ inline Src add(Src x, Src y) {
...
+inline Src mul(Src x, Src y, int32_t shift) {
+  assert(shift == 0); // shift has to be 0 if datatype is not int32_t
+  return emitc::mul(x, y);
+}
+
+template <>
+inline Tensor0D<int32_t> mul(Tensor0D<int32_t> x, Tensor0D<int32_t> y,
+                             int32_t shift) {
+  // Taken from
+  // https://git.mlplatform.org/tosa/reference_model.git/tree/reference_model/src/ops/ewise_binary.cc?id=df8626976df6c779bb30df9c5ceef689462109c0#n436
```
marbre, line 63: To which portion of you code does this apply? We may have to mark the taken over parts more clearly. Further, have you checked the license?

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,149 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+      for (int64_t d = numBroadcastDims; d < opOutputRank; d++) {
+        broadcastIndices.push_back(
+            mlir::IntegerAttr::get(rewriter.getIntegerType(64), d));
+      }
+
+      RankedTensorType ty =
+          RankedTensorType::get({static_cast<int64_t>(operandRank)},
+                                IntegerType::get(srcOp.getContext(), 64));
+      ArrayAttr broadcastArgs = rewriter.getArrayAttr(
+          DenseIntElementsAttr::get(ty, broadcastIndices));
```
marbre, line 148:
```suggestion
      RankedTensorType tensorType =
          RankedTensorType::get({static_cast<int64_t>(operandRank)},
                                IntegerType::get(srcOp.getContext(), 64));
      ArrayAttr broadcastArgs = rewriter.getArrayAttr(
          DenseIntElementsAttr::get(tensorType, broadcastIndices));
```

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,149 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+      // Replace the original operand with the result of the broadcast_in_dim
+      // operation
+      broadcastedOperands[i] = broadcastArg.getResult(0);
+    }
+  }
+  return broadcastedOperands;
+}
+
+template <typename SrcOp>
+class CallOpBroadcastableConversion : public OpConversionPattern<SrcOp> {
```
marbre, line 165: Maybe we should name this `BroadcastableCallOpConversion` instead of `CallOpBroadcastableConversion`? :)

@simon-camp @dajuro Your suggestions?

unittests/emitc_tosa.cpp
```diff
@@ -31,4 +34,43 @@ TEST(tosa, reciprocal) {
   EXPECT_THAT(result, Pointwise(FloatNear(EPSILON), excpected));
 }
 
+// Binary ops
```
marbre, line 37:
```suggestion
// Binary elementwise ops
```
To be consistent with your change above :)

marbre (changes requested):
Looks much better now. Still some nits.

In addition, do you have an idea how we can better test a broadcasted op? Can be done in a followup :)

include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,72 @@ inline Src add(Src x, Src y) {
   return emitc::add<Src>(x, y);
 }
 
+// MulOp
+template <typename Src>
+inline Src mul(Src x, Src y, int32_t shift) {
```
dajuro, line 54: Since you have written a `MulOpConversion` I suggest to drop `int32_t shift` and the assert completely: the compiler will be forced to choose the right function or fail miserable :)

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: No preference here.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,149 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
+  LogicalResult
+  matchAndRewrite(tosa::MulOp mulOp, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    StringRef funcName = "tosa::mul";
+    StringAttr callee = rewriter.getStringAttr(funcName);
+
+    SmallVector<Attribute, 1> args_;
+    args_.push_back(rewriter.getIndexAttr(0));
+    args_.push_back(rewriter.getIndexAttr(1));
+    args_.push_back(mulOp.shiftAttr());
```
dajuro, line 235: Just add 'shiftAttr' if needed. See my comment above: we can force the compiler to choose the right function.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -108,11 +108,149 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
   }
 };
 
+template <typename SrcOp>
+SmallVector<Value, 2>
+create_broadcast_op_if_needed(SrcOp &srcOp, ArrayRef<Value> operands,
```
dajuro, line 113: Do we need to call this in every conversion by hand?

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: I'm not quite sure why GitHub still shows the old code here. I already switched to post increment in this loop.

marbre: Have you pushed your changes?

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
marbre: Let's just keep it as it. We can still refactor in a followup if desired.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
marbre: I think this is just shared code. I assume we will not only have this in the broadcastable call op conversion.

include/emitc/emitc_tosa.h
```diff
@@ -49,6 +49,72 @@ inline Src add(Src x, Src y) {
...
+  assert(shift == 0); // shift has to be 0 if datatype is not int32_t
+  return emitc::mul(x, y);
+}
+
+template <>
+inline Tensor0D<int32_t> mul(Tensor0D<int32_t> x, Tensor0D<int32_t> y,
```
simon-camp, line 60: This won't work as expected, this function should be called for tensors with element type int32_t no matter what dimension.

simon-camp (changes requested):

include/emitc/emitc_tosa.h
OliverScherf: I commited your suggestion.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: This is in fact shared code. It is used in the `MulOpConversion` and `CallOpBroadcastableConversion`.

marbre: So we might revise the naming and check that is placed in the anonymous namespace (no need for a free function here), but in principle that's fine :)

marbre (approved):
LGTM if the pipeline has nothing to nag ;)

dajuro (approved):
LGTM

PR #67, opened 2021-03-02
Transform post increments to pre increments

marbre:


marbre:
We mainly use post increments in our header only implementations. Won't touch the other ones right now, as part of the mhlo stuff is moved to core ops.

simon-camp (approved):
LGTM.

PR #70, opened 2021-03-03
Implement basic ConstOp

marbre:
Our constant operation is intended to support opaque attributes as well
as our opaque type.

### DIFF
--- a/lib/Target/Cpp/TranslateToCpp.cpp
+++ b/lib/Target/Cpp/TranslateToCpp.cpp
@@ -60,11 +60,14 @@ static LogicalResult printConstOp(CppEmitter &emitter, ConstOp constOp) {
   auto value = constOp.value();
   bool emitBraces = value.isa<FloatAttr>() || value.isa<IntegerAttr>();
 
+  if (constOp.getType().dyn_cast<emitc::OpaqueType>()) {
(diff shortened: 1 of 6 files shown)


lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -49,6 +49,32 @@ static LogicalResult printConstantOp(CppEmitter &emitter,
...
+  // Add braces for number literals only to avoid double brace intialization for
+  // tensors.
+  auto value = constOp.value();
+  bool emitBraces = value.isa<FloatAttr>() || value.isa<IntegerAttr>();
+
+  if (emitBraces)
+    os << "{";
+
+  if (constOp.getType().dyn_cast<emitc::OpaqueType>())
+    os << " = ";
```
simon-camp, line 67: This goes wrong and prints something like `v1 { = ...}` when emitBraces is also true. We could emit an error in this case.

simon-camp (changes requested):
LGTM besides the comment.

simon-camp (approved):

lib/Target/Cpp/TranslateToCpp.cpp
marbre: I have addressed this with an additional commit. We should refactor a little larger in future, but I think that's fine for now.

PR #73, opened 2021-03-04
Add conversion for tosa conv2d op

dajuro:
This adds conversion, unittest and regression test for `tosa.conv2d` operation. Quantization is not supported.


include/emitc/emitc_tosa.h
```diff
@@ -98,6 +98,83 @@ inline Src mul(Src x, Src y, const int32_t shift) {
...
+// Conv2dOp
+template <typename Dest, typename Src, typename Weights>
+Dest conv2d(Src input, Weights weights, Tensor1D<int64_t, 4> padding,
+            Tensor1D<int64_t, 2> stride, Tensor1D<int64_t, 2> dilation) {
+  static_assert(is_tensor_of_dim<4, Src>::value,
+                "Expected 4 dimensional input");
+  static_assert(is_tensor_of_dim<4, Dest>::value,
+                "Expected 4 dimensional output");
+  static_assert(is_tensor_of_dim<4, Weights>::value,
+                "Expected 4 dimensional weights");
```
marbre, line 110: If we only support 4D Tensors here, why are we using a template?

marbre (commented):
@simon-camp You're more familiar with the conv in mhlo, so you might take a look as well.

include/emitc/emitc_tosa.h
dajuro: There is still `conv3d`. I hope that we can reuse this code/form a common base for both operations ;)

marbre: In that case, please add a comment why this is a template + static asserts. I think without this information in mind it is a little confusing and simply looks wrong.

include/emitc/emitc_tosa.h
```diff
@@ -98,6 +98,83 @@ inline Src mul(Src x, Src y, const int32_t shift) {
 }
 
 /// Other ops
+// Conv2dOp
```
marbre, line 101: NIT:
```suggestion
// Conv2DOp
```

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +81,65 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+    auto ctx = convOp.getContext();
+
+    StringRef funcName = "tosa::conv2d";
+    StringAttr callee = rewriter.getStringAttr(funcName);
+
+    // fail if quantization is requested
+    if (convOp.quantization_info().hasValue()) {
+      return convOp.emitError(
+          "Quantization of tosa.conv2d is currently not supported.");
+    }
```
simon-camp, line 109: Move this to the top of the function

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +81,65 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+
+    SmallVector<Attribute, 2> args_ = indexSequence(operands.size(), ctx);
+
+    args_.push_back(convOp.pad());
+    args_.push_back(convOp.stride());
+    args_.push_back(convOp.dilation());
+
+    ArrayAttr args = rewriter.getArrayAttr(args_);
```
simon-camp, line 117: Maybe it's easier to read if you specify the two operands directly via index attributes, i.e.
```c
ArrayAttr args = rewriter.getArrayAttr({
  rewriter.getIndexAttr(0),
  rewriter.getIndexAttr(1),
  convOp.pad(),
  convOp.stride(),
  convOp.dilation()
});
```
(untested)

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +81,65 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+  Conv2DOpConversion(MLIRContext *ctx) : OpConversionPattern(ctx) {}
+
+private:
+  LogicalResult
+  matchAndRewrite(tosa::Conv2DOp convOp, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    // tosa conv2D supports adding a bias after the actual convolution. We will
+    // split the convolution with bias into two operations: the convolution (as
+    // an emitc call op) and the addition of the bias (as an tosa.add op).
+    // Therefore we remove bias from the operands of the convolution.
```
simon-camp, line 96: Note: This is inconsistent with the implementation of the fully connected op which does the bias add in place I think.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +81,65 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+
+    args_.push_back(convOp.pad());
+    args_.push_back(convOp.stride());
+    args_.push_back(convOp.dilation());
+
+    ArrayAttr args = rewriter.getArrayAttr(args_);
+    ArrayAttr templateArgs =
+        rewriter.getArrayAttr({TypeAttr::get(convOp.getResult().getType()),
+                               TypeAttr::get(convOp.input().getType()),
+                               TypeAttr::get(convOp.weight().getType())});
```
simon-camp, line 121: Note: These two are optional, and are omitted for the fully connected op for readability reasons on the IR.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +81,65 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+    // prepare addOp
+    auto addOpOperands =
+        mlir::ValueRange({conv_op.getResult(0), convOp.bias()});
+
+    // create addOp
+    auto add_op = rewriter.create<tosa::AddOp>(convOp->getLoc(),
+                                               convOp.getType(), addOpOperands);
+
+    // rewrite convOp with two emitc CallOp
+    rewriter.replaceOp(convOp, {add_op.output()});
```
simon-camp, line 137: Can this be fused into a `rewriter.replaceWithNewOp`?

include/emitc/emitc_tosa.h
```diff
@@ -98,6 +98,83 @@ inline Src mul(Src x, Src y, const int32_t shift) {
...
+  assert(dilation[0] == 1);
+
+  const int N = input.dim(0);
+  const int H_IN = input.dim(1);
+  const int W_IN = input.dim(2);
+  const int C_IN = input.dim(3);
+
+  Dest output;
+
+  const int C_OUT = output.dim(0);
```
simon-camp, line 125: Must be output.dim(3)

simon-camp (changes requested):

simon-camp:
I made some comments, looks good otherwise.

include/emitc/emitc_tosa.h
dajuro: Beware: memory layout is different!

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: There will be a follow-up for the fully connected op.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: This is awesome!! Not sure if `clang-format` likes it :/

marbre: Give it a try. Otherwise we can disable clang-format here if it makes the code less readable :)

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: Thx! Good catch!

include/emitc/emitc_tosa.h
dajuro: We cannot use `Tensor4D` here because we don't know the size of the operands!

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: Yes it can! That is good to know!

include/emitc/emitc_tosa.h
OliverScherf: @marbre Using these static_assert generates more clear error messages than the compiler does IMO. We use the same technique in the MHLO Convolution implementation as well.

marbre: That comparison does not hold, as TOSA IR also restricts it to Tensor4D, whereas MHLO does not. The latter is the reason for inserting those into our MHLO header only implementation. But we're actually already discussing this *offline*, whatever this means in the current pandemic situation :laughing:

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: Just saw your comment: sadly `clang-format` doesn't like it.

OliverScherf:
LGTM! :+1: :+1:

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
marbre: That's why I meant you can turn it off here ;)

include/emitc/emitc_tosa.h
```diff
@@ -104,6 +104,78 @@ inline Src sub(Src x, Src y) {
...
+  // are [OC,KH,KW,IC] and output is [N,H,W,OC].
+  static_assert(is_tensor_of_dim<4, Src>::value,
+                "Expected 4 dimensional input");
+  static_assert(is_tensor_of_dim<4, Dest>::value,
+                "Expected 4 dimensional output");
+  static_assert(is_tensor_of_dim<4, Weights>::value,
+                "Expected 4 dimensional weights");
+
+  assert(stride[0] == 1);
+  assert(stride[0] == 1);
```
OliverScherf, line 122: `stride[1] == 1`

But why do you limit stride to be 1? Looking at the implementation strides are in fact supported.

include/emitc/emitc_tosa.h
```diff
@@ -104,6 +104,78 @@ inline Src sub(Src x, Src y) {
...
+
+  assert(dilation[0] == 1);
+  assert(dilation[0] == 1);
```
OliverScherf, line 125: `dilation[1] == 1`

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -72,6 +72,55 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+class Conv2DOpConversion : public OpConversionPattern<tosa::Conv2DOp> {
+
+public:
+  Conv2DOpConversion(MLIRContext *ctx) : OpConversionPattern(ctx) {}
+
+private:
+  LogicalResult
+  matchAndRewrite(tosa::Conv2DOp convOp, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    // fail if quantization is requested
```
OliverScherf, line 84: `// fails`

docs/tosa-op-coverage.md
```diff
@@ -20,5 +20,6 @@ The table below shows the supported TOSA ops.
 | mul                   | :heavy_check_mark: | |
 | sub                   | :heavy_check_mark: | |
 | **Other ops**
+| conv2d                | :white_check_mark: | Quantization not supported |
```
OliverScherf, line 23: Not only quantization but also dilated convs are not supported.

OliverScherf (changes requested):
Please look into this again, I added some comments :+1:

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
dajuro: I chose the imperative because I think it fits better here. It is supposed to fail! You have to read it with a drill sergeant voice in your mind ;)

include/emitc/emitc_tosa.h
dajuro: I took this from `emitc_mhlo.h` where the same checks are applied. Also we don't have tests for this. So it is not really supported and not tested. That's why there are this checks. Since we are not aiming for full feature compatibility in our first shot, I would leave it as it is.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: Sir yes sir! Ok fine by me

docs/tosa-op-coverage.md
dajuro: I updated the doc :+1:

OliverScherf (approved):

PR #74, opened 2021-03-04
Refactor to use `emitc.const` op

marbre:


simon-camp (approved):
LGTM. Nice cleanup :)

PR #78, opened 2021-03-04
Add conversion for tosa.matmul

OliverScherf:
This will also move the mhlo.dot op implementation to the emitc core ops, since we can reuse the implementation for the tosa.matmul op.

### DIFF
--- a/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
+++ b/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
@@ -121,7 +121,7 @@ class MatMulOpConversion : public OpConversionPattern<tosa::MatMulOp> {
   using OpConversionPattern<tosa::MatMulOp>::OpConversionPattern;
 
 public:
-  MatMulOpConversion(MLIRContext *ctx, StringRef funcName)
+  MatMulOpConversion(MLIRContext *ctx)
       : OpConversionPattern<tosa::MatMulOp>(ctx) {}
 
 private:
@@ -141,7 +141,7 @@ private:
     ArrayAttr args;
     ArrayAttr templateArgs;
 
-    rewriter.replaceOpWithNewOp<emitc::CallOp>(matMulOp, type, callee, args,
+    rewriter.replaceOpWithNewOp<emitc::CallOp>(matMulOp, matMulOp.getType(), callee, args,
                                                templateArgs, operands);
     return success();
   }
@@ -352,7 +352,7 @@ void populateTosaToEmitcPatterns(MLIRContext *ctx,
 
   // Other ops
   patterns.insert<FullyConnectedOpConversion>(ctx, "tosa::fully_connected");
-  patterns.insert<MatMulOpConversion>(ctx, "tosa::matmul");
+  patterns.insert<MatMulOpConversion>(ctx);
 }
 
 namespace {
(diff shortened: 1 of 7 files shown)


lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -117,6 +117,36 @@ class FullyConnectedOpConversion
...
+
+    StringRef funcName = "tosa::matmul";
+    StringAttr callee = rewriter.getStringAttr(funcName);
+
+    Type type = matMulOp.getType();
+
+    ArrayAttr args;
+    ArrayAttr templateArgs;
+
+    rewriter.replaceOpWithNewOp<emitc::CallOp>(matMulOp, type, callee, args,
```
marbre, line 144: I suggest to directly pass the type here
```suggestion
    rewriter.replaceOpWithNewOp<emitc::CallOp>(matMulOp, matMulOp.getType(), callee, args,
```
and drop it above. That's the way we do it in most conversions.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -117,6 +117,36 @@ class FullyConnectedOpConversion
   }
 };
 
+class MatMulOpConversion : public OpConversionPattern<tosa::MatMulOp> {
+  using OpConversionPattern<tosa::MatMulOp>::OpConversionPattern;
+
+public:
+  MatMulOpConversion(MLIRContext *ctx, StringRef funcName)
```
marbre, line 124:
```suggestion
  MatMulOpConversion(MLIRContext *ctx)
```
No need to pass `funcName`, you hardcode it anyway.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -322,6 +352,7 @@ void populateTosaToEmitcPatterns(MLIRContext *ctx,
 
   // Other ops
   patterns.insert<FullyConnectedOpConversion>(ctx, "tosa::fully_connected");
+  patterns.insert<MatMulOpConversion>(ctx, "tosa::matmul");
```
marbre, line 355:
```suggestion
  patterns.insert<MatMulOpConversion>(ctx);
```

test/Conversion/tosa-to-emitc.mlir
```diff
@@ -129,3 +129,11 @@ func @test_fully_connected(%arg0: tensor<14x19xf32>, %arg1: tensor<19x28xf32>, %
   %0 = "tosa.fully_connected"(%arg0, %arg1, %arg2) : (tensor<14x19xf32>, tensor<19x28xf32>, tensor<28xf32>) -> tensor<14x28xf32>
   return %0 : tensor<14x28xf32>
 }
+
+
+// MatMulOp
+func @test_matmul(%arg0: tensor<14x19xf32>, %arg1: tensor<19x28xf32>) -> tensor<14x28xf32> {
+  // CHECK: emitc.call "tosa::matmul"(%arg0, %arg1) : (tensor<14x19xf32>, tensor<19x28xf32>) -> tensor<14x28xf32>
+  %0 = "tosa.matmul"(%arg0, %arg1) : (tensor<14x19xf32>, tensor<19x28xf32>) -> tensor<14x28xf32>
```
marbre, line 137: Can we really handle it when tensors have different num. elements?

marbre (changes requested):

test/Conversion/tosa-to-emitc.mlir
marbre: Okay, discussed this and it's fine :)

marbre (approved):
Nice :)

PR #81, opened 2021-03-08
Add unittests for strided Conv2d

OliverScherf:
This verifies that the TOSA and MHLO convolution implementation support strided convolutions by adding unittests.

### DIFF
--- a/include/emitc/emitc_tosa.h
+++ b/include/emitc/emitc_tosa.h
@@ -120,6 +120,9 @@ Dest conv2d(Src input, Weights weights, Tensor1D<int64_t, 4> padding,
   static_assert(is_tensor_of_dim<4, Weights>::value,
                 "Expected 4 dimensional weights");
 
+  assert(stride[0] > 0);
+  assert(stride[1] > 0);
+
   assert(dilation[0] == 1);
   assert(dilation[1] == 1);
 
--- a/include/emitc/emitc_tosa.h
+++ b/include/emitc/emitc_tosa.h
@@ -120,9 +120,6 @@ Dest conv2d(Src input, Weights weights, Tensor1D<int64_t, 4> padding,
   static_assert(is_tensor_of_dim<4, Weights>::value,
                 "Expected 4 dimensional weights");
 
-  assert(stride[0] == 1);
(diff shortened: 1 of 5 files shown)


include/emitc/emitc_tosa.h
```diff
@@ -120,9 +120,6 @@ Dest conv2d(Src input, Weights weights, Tensor1D<int64_t, 4> padding,
   static_assert(is_tensor_of_dim<4, Weights>::value,
                 "Expected 4 dimensional weights");
 
-  assert(stride[0] == 1);
-  assert(stride[1] == 1);
```
simon-camp, line 124: Maybe
```c
assert(stride[0] > 0);
assert(stride[1] > 0);
```

to be on the safe side?

simon-camp (changes requested):

include/emitc/emitc_tosa.h
marbre: Side note, not relevant to this PR: Should the base type of `stride` be an unsigned int?

simon-camp: These are signless integers on the mlir side, which we emit as signed integers in C++. I'm not sure if we want to have implicit conversions going on there.

OliverScherf: @simon-camp I dont think this check is needed, as this only would occur with invalid IR?
@marbre Speaking about tosa.conv2d: TOSA MLIR Dialect specifies strides as 64 bit integer array. Logically you're right.

marbre: > These are signless integers on the mlir side, which we emit as signed integers in C++. I'm not sure if we want to have implicit conversions going on there.

Yeah, that was the question. I am not sure either, that's why I asked.

dajuro (dismissed):
Tests look good :+1:

include/emitc/emitc_tosa.h
OliverScherf: I added assert statements to check that the strides are positive.

dajuro (approved):

simon-camp (approved):

PR #83, opened 2021-03-08
Don't emit assignment operator for empty consts

marbre:


lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -66,19 +66,26 @@ static LogicalResult printConstOp(CppEmitter &emitter, ConstOp constOp) {
...
   bool emitBraces = value.isa<FloatAttr>() || value.isa<IntegerAttr>();
 
-  // Never emit curly braces if
-  // - restricted to C or
-  // - assigning to an EmitC opaque type
-  if (emitter.restrictedToC() ||
-      constOp.getType().dyn_cast<emitc::OpaqueType>()) {
+  // Add braces for non empty StringAttr
+  if (value.isa<StringAttr>()) {
+    auto sAttr = value.dyn_cast<StringAttr>();
```
simon-camp, line 71: Maybe like this?
```suggestion
  if (auto sAttr = value.dyn_cast<StringAttr>()) {
```

test/Target/cpp-calls.mlir
```diff
@@ -88,10 +88,15 @@ func @get_address_of(%arg0: i32) -> !emitc.opaque<"int32_t*"> {
...
 }
 
+// CHECK: void emitc_constant()
 func @emitc_constant() {
-  // CHECK: int32_t [[V1]]{42};
-  %1 = "emitc.const"(){value = 42 : i32} : () -> i32
-  // CHECK: int32_t* [[V2]] = nullptr;
-  %2 = "emitc.const"(){value = "nullptr" : !emitc.opaque<"..">} : () -> !emitc.opaque<"int32_t*">
+  // CHECK: int32_t [[V0:[^ ]*]];
+  %c0 = "emitc.const"(){value = "" : !emitc.opaque<" ">} : () -> i32
```
simon-camp, line 94: We should add a verfication that the op result type (i32 here) and value attribute type (!emitc.opaque<" ">) match. This is also done for the std constant op.

```c++
if (!value.getType().isa<NoneType>() && type != value.getType())
    return op.emitOpError() << "requires attribute's type (" << value.getType()
                            << ") to match op's return type (" << type << ")";
```

I'd suggest to make the types the same on the lines touched by this PR and fix the rest in a follow up.

```suggestion
  %c0 = "emitc.const"(){value = "" : !emitc.opaque<"int32_t">} : () -> !emitc.opaque<"int32_t">
```
(untested)

simon-camp (changes requested):
LGTM, just one suggestion and a comment which we can address in a follow up.

test/Target/cpp-calls.mlir
marbre: I changed this to `{value = "" : i32} : () -> i32` instead.

simon-camp: Sure, somehow I thought that wouldn't verify right now :)

marbre: Right now, we don't verify anything, so you can actually do whatever comes to your mind :exploding_head:

PR #88, opened 2021-03-10
Add conversion for tosa.reluN

OliverScherf:
This also moves the mhlo.clamp op implementation to the emitc core ops, so that we can reuse the implementation for tosa.reluN.

### DIFF
--- a/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
+++ b/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
@@ -241,7 +241,8 @@ private:
     SmallVector<Attribute, 2> args_;
     args_.push_back(rewriter.getIndexAttr(0));
 
-    // Determine to which max value we have to clamp to
+    // Since tosa.reluN has two max attribute types for float and integer
+    // values, we have to determine to which max attribute we have to clamp to
     auto operandType =
         operands[0].getType().cast<RankedTensorType>().getElementType();
     if (operandType.isSignedInteger() || operandType.isSignlessInteger() ||
--- a/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
+++ b/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
@@ -243,14 +243,11 @@ private:
 
     // Since tosa.reluN has two max attribute types for float and integer
     // values, we have to determine to which max attribute we have to clamp to
-    auto operandType =
+    auto elementType =
         operands[0].getType().cast<RankedTensorType>().getElementType();
-    if (operandType.isSignedInteger() || operandType.isSignlessInteger() ||
-        operandType.isUnsignedInteger()) {
+    if (elementType.isa<IntegerType>()) {
       args_.push_back(reluNOp.max_intAttr());
-    } else if (operandType.isF16() || operandType.isF32() ||
-               operandType.isF64() || operandType.isF80() ||
-               operandType.isF128()) {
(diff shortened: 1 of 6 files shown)


lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -223,6 +223,48 @@ class MatMulOpConversion : public OpConversionPattern<tosa::MatMulOp> {
...
+
+    // Determine to which max value we have to clamp to
+    auto operandType =
+        operands[0].getType().cast<RankedTensorType>().getElementType();
+    if (operandType.isSignedInteger() || operandType.isSignlessInteger() ||
+        operandType.isUnsignedInteger()) {
+      args_.push_back(reluNOp.max_intAttr());
+    } else if (operandType.isF16() || operandType.isF32() ||
+               operandType.isF64() || operandType.isF80() ||
+               operandType.isF128()) {
```
simon-camp, line 253: I think you can use `isa<FloatType>(operandType)` and `isa<IntegerType>(operandType)`
And maybe call the variable `elementType`.

simon-camp (changes requested):
I think the condition can be simplified, looks good otherwise.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: @simon-camp I was looking for something like that but couldnt find it! Thanks!

marbre: Sometimes it's worth to take a look into the printer :)
https://github.com/iml130/mlir-emitc/blob/8f4f56c2d5042dd4e770a1ac9416f989e0480662/lib/Target/Cpp/TranslateToCpp.cpp#L37-L38

dajuro (approved):
LGTM :+1:

simon-camp (approved):

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -223,6 +223,46 @@ class MatMulOpConversion : public OpConversionPattern<tosa::MatMulOp> {
...
+    StringRef funcName = "tosa::reluN";
+    StringAttr callee = rewriter.getStringAttr(funcName);
+
+    SmallVector<Attribute, 2> args_;
+    args_.push_back(rewriter.getIndexAttr(0));
+
+    // Since tosa.reluN has two max attribute types for float and integer
+    // values, we have to determine to which max attribute we have to clamp to
+    auto elementType =
+        operands[0].getType().cast<RankedTensorType>().getElementType();
```
marbre, line 247: You may need to check if this is a ranked Tensor. I see two options:

```c++
    if (auto tensorType = operands[0].getType().dyn_cast<RankedTensorType>())
      auto elementType = tensorType.getElementType();
      
      ...
      
    } else {
      return failure();
    }
```

but I would probably go for a more general check. Instead one could go for

```suggestion
    auto tensorType = operands[0].getType().dyn_cast<TensorType>()
    if (!tensorType.hasRank())
      return failure();
    auto elementType = tensorType.getElementType();
```

marbre (changes requested):

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
marbre: We won't check this here as discussed internally.

PR #89, opened 2021-03-10
Add cli option to forward declare variables

simon-camp:


lib/Target/Cpp/TranslationFlags.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "llvm/Support/CommandLine.h"
+
+namespace mlir {
+namespace emitc {
+
+static llvm::cl::opt<bool> forwardDeclareFlag{
+    "forward-declare-variables",
+    llvm::cl::desc(
+        "Forward declares all variables at the top of emitted functions."),
```
marbre, line 20:
```suggestion
        "Forward declare all variables emitted from emitc"),
```

"at the top .. functions" might be misleading?

marbre (changes requested):
One minor nit, looks good beside that.

marbre (approved):
Ready to be merged, as soon as clang-format is happy :)

marbre (approved):

PR #90, opened 2021-03-10
Refactor CppEmitter::emitType

marbre:
Harmonizes the naming of variables storing type information and
fixes/improves checking tensor types. Till now we only checked if a
tensor is ranked, but not if is has a static shape.


OliverScherf (approved):
LGTM :+1:

PR #92, opened 2021-03-11
Fix CallOpBroadcastableConversion

OliverScherf:
This will fix a bug introduced with the CallOpBroadcastableConversion. The create CallOp to broadcast_in_dim op was missing the index attribute for the operand.

Furthermore this adds a unittest to simulate the behaviour of a broadcastable tosa op.

### DIFF
--- a/unittests/emitc_tosa.cpp
+++ b/unittests/emitc_tosa.cpp
@@ -88,8 +88,8 @@ TEST(tosa, mul) {
 
 TEST(tosa, broadcastable_op) {
   // In the CallOpBroadcastableConversion ops where the tensor shape of the
-  // operands don't match, a broadcast_in_dim op is inserted. This unittest is
-  // to verify that procedure.
+  // operands don't match, a broadcast_in_dim op is inserted. This unittest
+  // verifies that procedure.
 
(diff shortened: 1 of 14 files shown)


unittests/emitc_tosa.cpp
```diff
@@ -86,6 +86,32 @@ TEST(tosa, mul) {
   EXPECT_THAT(lambda_1d_int_shift(), Pointwise(Eq(), {0, 1, 2, 4}));
 }
 
+TEST(tosa, broadcastable_op) {
+  // In the CallOpBroadcastableConversion ops where the tensor shape of the
+  // operands don't match, a broadcast_in_dim op is inserted. This unittest is
+  // to verify that procedure.
```
simon-camp, line 92:
```suggestion
  // operands don't match, a broadcast_in_dim op is inserted. This unittest
  // verifies that procedure.
```

simon-camp (changes requested):
Nice catch :)

simon-camp (approved):
Please fix the last sentence in the PR description (singular op at the end) :)

marbre (approved):

PR #93, opened 2021-03-11
Add emitVariableDeclaration method

simon-camp:
Add a new helper method which emits the type and a new variable name for an OpResult.


marbre (approved):
Awesome! :)

PR #94, opened 2021-03-11
Fix pass pipeline in e2e script

simon-camp:


simon-camp:
We don't use this script in the github workflow for Mobilenet because we cache the intermediate results. Maybe a single layer test model can be run fast enough without caching, so that we can call this script in the test pipeline.

marbre (approved):

PR #95, opened 2021-03-11
Fix tosa.reluN conversion

OliverScherf:
Fixes the case, where the bit width of the operand does not match the one from the max attributes.


dajuro (dismissed):
LGTM :+1:

marbre (approved):

PR #99, opened 2021-03-15
Add TOSA e2e test

OliverScherf:
This adds an e2e test for TOSA. For pipeline purposes this adds the MobileNetV2 as TOSA MLIR as well as a test program to verify the results of the generated emitC C++ code.


marbre (approved):
Awesome :)

PR #100, opened 2021-03-15
Add Doxygen comments and cleanup existing comments

marbre:


lib/Dialect/EmitC/Conversion/MHLOToEmitC.cpp
```diff
@@ -41,6 +41,7 @@ SmallVector<Attribute, 2> indexSequence(int64_t n, MLIRContext *ctx) {
       }));
 }
 
+/// Convert `mhlo.constant` into an `emic.const` operation.
```
OliverScherf, line 44: This is a copy pasted typo, search for `emic` in your changes, it appears in multiple places

```suggestion
/// Convert `mhlo.constant` into an `emitc.const` operation.
```

lib/Dialect/EmitC/Conversion/MHLOToEmitC.cpp
```diff
@@ -608,7 +621,7 @@ struct ConvertMhloToEmitCPass
     // MHLO region ops
     target.addIllegalOp<mhlo::ReduceOp,
                         mhlo::ReturnOp>();
-    // other MHLO ops
+    // Other MHLO ops
```
OliverScherf, line 624:
```suggestion
    // Other MHLO ops.
```

OliverScherf (changes requested):

lib/Dialect/EmitC/Conversion/MHLOToEmitC.cpp
marbre: Well that is what we have discussed this morning and therefore I didn't adjust all of these *short* comments.

lib/Dialect/EmitC/Conversion/MHLOToEmitC.cpp
marbre: Good catch! All done :)

lib/Dialect/EmitC/Conversion/MHLOToEmitC.cpp
OliverScherf: Yeah I know we didn't found a general solution to this, but I suggested the change because you ended all the other // MHLO ops with a dot as well.

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -331,7 +339,7 @@ createBroadcastOpIfNeeded(SrcOp &srcOp, ArrayRef<Value> operands,
     auto operandShape = operand.getType().cast<RankedTensorType>().getShape();
     auto operandRank = operand.getType().cast<RankedTensorType>().getRank();
 
-    // Insert a broadcast_in_dim Operation if shape of operands don't match
+    // Insert a broadcast_in_dim Operation if shape of operands don't match.
```
OliverScherf, line 342:
```suggestion
    // Insert a broadcast_in_dim operation if shape of operands don't match.
```

OliverScherf (changes requested):
I found some little NIT :)

PR #101, opened 2021-03-15
Implement support for forward declarations of variables

simon-camp:


include/emitc/Target/Cpp.h
```diff
@@ -86,6 +86,9 @@ struct CppEmitter {
   /// the array size.
   LogicalResult emitTupleType(ArrayRef<Type> types);
 
+  /// Emits an assignment for a variable which has been declared previously.
+  LogicalResult emitVariableAssignment(OpResult result);
+
   /// Emits a variable declaration for a result of an operation.
   LogicalResult emitVariableDeclaration(OpResult result,
                                         bool trailingSemicolon);
```
marbre, line 94: Do we want to reintroduce the default `= true` here, if it is removed from `emitOperation`?

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -28,26 +28,34 @@ using llvm::formatv;
 static LogicalResult printConstantOp(CppEmitter &emitter,
                                      ConstantOp constantOp) {
   auto &os = emitter.ostream();
-  if (failed(emitter.emitVariableDeclaration(
-          constantOp.getOperation()->getResult(0),
-          /*trailingSemicolon=*/false)))
-    return failure();
+  auto result = constantOp.getOperation()->getResult(0);
+  auto value = constantOp.value();
```
marbre, line 32: Hm, any chance to not loose the check?

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -28,26 +28,34 @@ using llvm::formatv;
...
+  // Emit an assignment if variables are forward declared.
+  if (emitter.forwardDeclaredVariables()) {
+    if (failed(emitter.emitVariableAssignment(result)))
+      return failure();
+  } else {
+    if (failed(emitter.emitVariableDeclaration(result,
+                                               /*trailingSemicolon=*/false)))
+      return failure();
+    if (emitter.restrictedToC())
+      os << " = ";
```
marbre, line 52: Why not
```suggestion
    if (emitter.restrictedToC()) {
      os << " = ";
    } else {
      os << "{";
    }
```
Makes somehow more sense to me to put this here and to remove below. At least at this late hour ;)

marbre (changes requested):
Some first suggestions. Still need to go through the if and for op printer.

include/emitc/Target/Cpp.h
simon-camp: I'd vote for no. Having to think about this for each operation is a good thing.

simon-camp:
I think this is feature complete for now. Tests for `emitc.call, emitc.const,emitc.getaddressof, !emitc.opaque`, need to be added/refactored, so they are run on all (applicable) target combinations.

include/emitc/Target/Cpp.h
marbre: I am fine with this :+1:

marbre (approved):
Awesome work Simon!

PR #106, opened 2021-03-18
Don't emit empty template arguments

OliverScherf:
This cleans up the IR by not emitting empty template args.


marbre (approved):
Nice cleanup :)

PR #107, opened 2021-03-18
Expand and refactor tests

marbre:
We now explicitly check that were piped to emitc-translate before and not checked separately. Furthermore, the MHLO to EmitC tests are expanded (check inputs and outputs as well as arguments and template arguments). The latter also revealed that some non empty template arguments were not checked before.

### DIFF
--- a/test/Conversion/std-to-emitc.mlir
+++ b/test/Conversion/std-to-emitc.mlir
@@ -1,20 +1,30 @@
-// RUN: emitc-opt -convert-std-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s
+// RUN: emitc-opt -convert-std-to-emitc %s | FileCheck %s
+// RUN: emitc-opt -convert-std-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s -check-prefix=CPP
 
-// CHECK: Tensor<size_t, 2> std_index_cast(Tensor<size_t> v1, Tensor<int32_t, 2> v2, Tensor<int32_t, 2, 2> v3)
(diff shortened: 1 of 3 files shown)


test/Conversion/std-to-emitc.mlir
```diff
@@ -1,20 +1,30 @@
-// RUN: emitc-opt -convert-std-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s
+// RUN: emitc-opt -convert-std-to-emitc %s
+// RUN: emitc-opt -convert-std-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s -check-prefix=CPP
```
OliverScherf, line 2: These two line are intended?

test/Conversion/tensor-to-emitc.mlir
```diff
@@ -1,14 +1,22 @@
-// RUN: emitc-opt -convert-tensor-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s
+// RUN: emitc-opt -convert-tensor-to-emitc %s
+// RUN: emitc-opt -convert-tensor-to-emitc %s | emitc-translate --mlir-to-cpp | FileCheck %s -check-prefix=CPP
```
OliverScherf, line 2: Same question

OliverScherf (commented):
LGTM, I just added one question.

test/Conversion/std-to-emitc.mlir
marbre: Yes, to run two separate tests.

marbre:
Rebased onto main.

marbre:
Added missing `FileCheck` calls.

simon-camp (approved):
LGTM :)

PR #108, opened 2021-03-18
Rename translate() to convert()

marbre:
Referring to the MLIR glossary, this is a conversion and not a
translation.


dajuro (approved):
:+1:

PR #109, opened 2021-03-22
Disallow template args if translating to C

OliverScherf:
Also adds a new test file for testing restricted emitC features when translating to C.


lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -134,6 +134,10 @@ static LogicalResult printCallOp(CppEmitter &emitter, emitc::CallOp callOp) {
   };
 
   if (callOp.template_args()) {
+    if (emitter.restrictedToC()) {
+      return op.emitOpError()
+             << "template arguments are invalid if translating to C";
```
marbre, line 139:
```suggestion
             << "template arguments are not supported if emitting C";
```

test/Target/invalid-c-translate.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: emitc-translate -split-input-file -mlir-to-c -verify-diagnostics %s
+
+func @func_template() {
+  // expected-error @+1 {{'emitc.call' op template arguments are invalid if translating to C}}
```
marbre, line 4:
```suggestion
  // expected-error @+1 {{'emitc.call' op template arguments are not supported if emitting C}}
```

marbre (changes requested):
Only two minor suggestions. In addition, may we want to rename `invalid-c-translate.mlir` to `invalid-c-translation.mlir` or simply to `invalid-c`? :)

marbre (approved):
:+1:

PR #110, opened 2021-03-22
Implement control flow via labels and gotos.

simon-camp:


lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -94,6 +94,71 @@ static LogicalResult printConstantOp(CppEmitter &emitter,
...
+static LogicalResult printCondBranchOp(CppEmitter &emitter,
+                                       mlir::CondBranchOp condBranchOp) {
+  auto &os = emitter.ostream();
+  Block &trueSuccessor = *condBranchOp.getTrueDest();
+  Block &falseSuccessor = *condBranchOp.getFalseDest();
+
+  os << "if (" << emitter.getOrCreateName(condBranchOp.getCondition())
+     << ") {\n";
+
+  // true case
```
marbre, line 127:
```suggestion
  // If condition is true.

File: lib/Target/Cpp/TranslateToCpp.cpp [hunk 2]
```diff
@@ -94,6 +94,71 @@ static LogicalResult printConstantOp(CppEmitter &emitter,
...
+  }
+
+  os << "goto ";
+  if (!(emitter.hasBlockLabel(trueSuccessor))) {
+    return condBranchOp.emitOpError()
+           << "Unable to find label for successor block";
+  }
+  os << emitter.getOrCreateName(trueSuccessor) << ";\n";
+  os << "} else {\n";
+  // false case
```
marbre, line 143:
```suggestion
  // If condition is false.
```

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -416,11 +485,42 @@ static LogicalResult printFunction(CppEmitter &emitter, FuncOp functionOp) {
...
+  for (auto &block : blocks) {
+    // Only print a label if there is more than one block.
+    if (blocks.size() > 1) {
+      if (failed(emitter.emitLabel(block))) {
+        return functionOp.emitOpError()
+               << "Unable to print label for basic block";
+      }
+    }
+    for (Operation &op : block.getOperations()) {
+      // Don't print additional semicolons after these operations.
```
marbre, line 516: Nice!

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -651,23 +762,38 @@ LogicalResult CppEmitter::emitAssignPrefix(Operation &op) {
...
+    return block.getParentOp()->emitError("Label for block not found.");
+  }
+  os << getOrCreateName(block) << ":\n";
+  return success();
+}
+
 static LogicalResult printOperation(CppEmitter &emitter, Operation &op) {
-  if (auto callOp = dyn_cast<mlir::CallOp>(op))
-    return printCallOp(emitter, callOp);
+  // emitc ops
```
marbre, line 774:
```suggestion
  // EmitC ops.
```

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -651,23 +762,38 @@ LogicalResult CppEmitter::emitAssignPrefix(Operation &op) {
...
   if (auto callOp = dyn_cast<emitc::CallOp>(op))
     return printCallOp(emitter, callOp);
+  if (auto constOp = dyn_cast<emitc::ConstOp>(op))
+    return printConstantOp(emitter, constOp);
```
marbre, line 778: Do we want to use the order in the IR definition or do we want to sort alphabetically?

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -651,23 +762,38 @@ LogicalResult CppEmitter::emitAssignPrefix(Operation &op) {
...
   if (auto ifOp = dyn_cast<emitc::IfOp>(op))
     return printIfOp(emitter, ifOp);
-  if (auto yieldOp = dyn_cast<emitc::YieldOp>(op))
-    return printYieldOp(emitter, yieldOp);
   if (auto forOp = dyn_cast<emitc::ForOp>(op))
     return printForOp(emitter, forOp);
+  if (auto yieldOp = dyn_cast<emitc::YieldOp>(op))
+    return printYieldOp(emitter, yieldOp);
+  // std ops
+  // control flow
```
marbre, line 788:
```suggestion
  // Standard ops.
```

IMHO no need to add a comment on *control flow* if it's just a single op/printer function.

marbre (commented):
LGTM, but will test locally before approving. And of course, some remarks ;)

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -375,9 +440,13 @@ static LogicalResult printModule(CppEmitter &emitter, ModuleOp moduleOp) {
...
 static LogicalResult printFunction(CppEmitter &emitter, FuncOp functionOp) {
-  auto &blocks = functionOp.getBlocks();
-  if (blocks.size() != 1)
-    return functionOp.emitOpError() << "only single block functions supported";
+  // We need to forward-declare variables if the function has multiple blocks.
+  if (!emitter.forwardDeclaredVariables() &&
+      functionOp.getBlocks().size() > 1) {
+    return functionOp.emitOpError()
+           << "Need to enable variable forward declaration when emitting code "
+              "for function with multiple blocks";
```
marbre, line 448: I think we could improve this error message, as it gets you

```
test/Target/control_flow.mlir:6:1: error: 'func' op Need to enable variable forward declaration when emitting code for function with multiple blocks
```

in the case you don't pass `-forward-declare-variables`.

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -435,9 +535,16 @@ CppEmitter::CppEmitter(raw_ostream &os, bool restrictToC,
...
 StringRef CppEmitter::getOrCreateName(Value val) {
-  if (!mapper.count(val))
-    mapper.insert(val, formatv("v{0}", ++valueInScopeCount.top()));
-  return *mapper.begin(val);
+  if (!valMapper.count(val))
+    valMapper.insert(val, formatv("v{0}", ++valueInScopeCount.top()));
+  return *valMapper.begin(val);
+}
+
+/// Return the existing or a new name for a Block*.
```
marbre, line 543:
```suggestion
/// Return the existing or a new name for a block.
```

Is "Block" with capital letter + asterisk intended?

simon-camp: The capital letter is intended as it refers to the data type. The asterisk is a typo in the original comment on the Value overload of the function. Both should match now.

marbre (approved):

PR #111, opened 2021-03-23
Start label counting at 1 for every function

simon-camp:
Before labels were named based on a module wide counter.


marbre (dismissed):
Side note, isn't the expected test output identical for C and C++? Is it useful to test twice if the printer pass doesn't differs?

simon-camp:
Yes it should be identical (and we test this here), but we could in principle lower these two differently in the future. Then we already have a test in place. We could use the same check-prefix though.

marbre:
We can leave it as is and cleanup in a followup if we really want to ;)

marbre (approved):

PR #114, opened 2021-03-23
[nfc] Cleanup

marbre:


lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -153,7 +153,7 @@ class GenericConvOpConversion : public OpConversionPattern<SrcOp> {
     ArrayAttr templateArgs =
         rewriter.getArrayAttr({TypeAttr::get(convOp.getResult().getType())});
 
-    // Create convOp
+    // Create convOp.
```
OliverScherf, line 156:
```suggestion
    // Create conv op.
```

OliverScherf (changes requested):
One little thing, otherwise LGTM!

OliverScherf (approved):

PR #116, opened 2021-03-31
Add conversions for tosa.maximum and tosa.minimum

OliverScherf:


dajuro (approved):
LGTM :+1:  Out of curiosity: do we have/does TOSA have non-binary max/min ops?

OliverScherf:
> LGTM +1 Out of curiosity: do we have/does TOSA have non-binary max/min ops?

There is tosa.argmax, which returns the index of the maximum value.

dajuro (approved):
Beautiful :+1:

PR #118, opened 2021-03-31
Add conversion for tosa.negate

OliverScherf:


lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -233,6 +233,35 @@ class MatMulOpConversion : public OpConversionPattern<tosa::MatMulOp> {
...
+
+public:
+  NegateOpConversion(MLIRContext *ctx)
+      : OpConversionPattern<tosa::NegateOp>(ctx) {}
+
+private:
+  LogicalResult
+  matchAndRewrite(tosa::NegateOp negateOp, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (negateOp.quantization_info().hasValue()) {
```
dajuro, line 248: Is this the cause for the custom OpConversion?

dajuro (approved):
Nice one!

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
OliverScherf: Yes, it's solely to forbid the quantization.

PR #121, opened 2021-04-09
Remove trailing comma from attributes

marbre:
Makes sure to remove if trailing commas are present after stripping
`tf._input_shapes` from the attributes (it might be that it was the last
attribute).


marbre:
@OliverScherf please check if this resolves the problems you faced when using TF 2.5.

OliverScherf (approved):
Good job :rocket:

PR #126, opened 2021-04-14
Emit more meaningful translate errors

OliverScherf:
This adds more meaningful errors messages to emitc-translate as well as tests for the new (and already existing) error messages.


marbre:
Instead of passing operation references, one can pass operation pointers and use `op.getOperation()`. That's the way we used before. Having discusses this with @simon-camp, we agreed that this should be refactored.

test/Target/invalid.mlir
```diff
@@ -8,3 +8,42 @@ func @multiple_blocks() {
...
+func @unsupported_float_type(%arg0 : f80) {
+  return
+}
+
+// -----
+
+// expected-error@+1 {{cannot emit type 'memref<100xf32>'}}
+func @memref_type(%arg0 : memref<100xf32>) {
+  return
+}
```
simon-camp, line 47: Add a copy of this  with type `vector<100xf32>`

test/Target/invalid-cpp.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: emitc-translate -split-input-file -mlir-to-cpp -verify-diagnostics %s
+
+// expected-error@+1 {{cannot emit tensor type with non static shape}}
+func @non_static_shape(%arg0 : tensor<?xf32>) {
+  return
+}
```
simon-camp, line 6: Add a test for unranked tensors `tensor<*xf32>`

simon-camp (changes requested):
Two tests you might want to add. Otherwise looks good besides what @marbre already suggested.

test/Target/invalid-cpp.mlir
OliverScherf: And so it shall be.

test/Target/invalid.mlir
OliverScherf: And so it shall be.

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -659,21 +664,21 @@ LogicalResult CppEmitter::emitAttribute(Attribute attr) {
...
     }
     os << sAttr.getRootReference();
     return success();
   }
   if (auto type = attr.dyn_cast<TypeAttr>()) {
-    return emitType(type.getValue());
+    return emitType(op, type.getValue());
   }
-  return failure();
+  return op.emitError("cannot emit attribute of type ") << attr.getType();
```
marbre, line 675: Can we add a test for this or is this one of the untestable errors?

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -507,8 +513,7 @@ static LogicalResult printFunction(CppEmitter &emitter, FuncOp functionOp) {
     // Only print a label if there is more than one block.
     if (blocks.size() > 1) {
       if (failed(emitter.emitLabel(block))) {
-        return functionOp.emitOpError()
-               << "Unable to print label for basic block";
+        return failure();
```
marbre, line 516: Can you elaborate on this?

marbre (changes requested):
Some final remarks :)

lib/Target/Cpp/TranslateToCpp.cpp
OliverScherf: This code shouldn't be reached, unless we missed a possible attribute type. So currently there is no known way to produce and test this error.

test/Target/invalid.mlir
```diff
@@ -8,3 +8,47 @@ func @multiple_blocks() {
...
     return
 }
+
+// -----
+
+func @index_argument_out_of_range(%arg0 : f32, %arg1 : f32) {
+  // expected-error@+1 {{'emitc.call' op index argument is out of range}}
+  %0 = emitc.call "add"(%arg0, %arg1) {args = [1 : index, 2 : index]} : (f32, f32) -> f32
+  return
+}
```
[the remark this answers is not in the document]
OliverScherf: I removed this test from the PR.

lib/Target/Cpp/TranslateToCpp.cpp
marbre: Okay :)

lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -474,8 +480,7 @@ static LogicalResult printFunction(CppEmitter &emitter, FuncOp functionOp) {
           for (auto result : op->getResults()) {
             if (failed(emitter.emitVariableDeclaration(
                     result, /*trailingSemicolon=*/true))) {
-              return WalkResult(
-                  op->emitError("Unable to declare result variable for op"));
+              return WalkResult(failure());
```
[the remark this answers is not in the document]
OliverScherf: As errors are getting concatenated I added this message to the code again.

marbre (approved):

marbre:
Nice improvement before starting into the weekend :partying_face:

PR #131, opened 2021-04-23
Reorder arguments passed to `add_mlir_doc`

marbre:
The `add_mlir_doc` CMake macro has changed upstream with commit
llvm/llvm-project@90ae4d90.


marbre:
@dajuro it would be nice if you take over the review as Simon isn't available.

dajuro (approved):
Looks good :+1: I was confused there for a minute because the  [MLIR doc](https://mlir.llvm.org/docs/Tutorials/CreatingADialect/#cmake-best-practices) is still oudated...

marbre:
Excellent catch, will update those ASAP in upstream.

PR #133, opened 2021-04-27
Drop GetAddressOfOp in favor of ApplyOp

marbre:
This replaces the `emitc.getaddressof` op in favor of the newly added
`emitc.apply`. The `emitc.apply` op allows to specify the operator to
apply (the applicable operator). This not only allows a "get address"
using the & operator, as it also enables a "get content" via the *
operator.

### DIFF
--- a/lib/Dialect/EmitC/IR/EmitCDialect.cpp
+++ b/lib/Dialect/EmitC/IR/EmitCDialect.cpp
@@ -51,6 +51,25 @@ void emitc::buildTerminatedBody(OpBuilder &builder, Location loc) {
 }
 
 //===----------------------------------------------------------------------===//
+// ApplyOp
+//===----------------------------------------------------------------------===//
+
+static LogicalResult verify(mlir::emitc::ApplyOp op) {
+  StringRef applicableOperator = op.applicableOperator();
+
+  // Applicable operator must not be empty.
+  if (applicableOperator.empty()) {
+    return op.emitOpError("applicable operator must not be empty");
+  }
+
+  // Only `*` and `&` are supported.
+  if (!applicableOperator.equals("&") && !applicableOperator.equals("*"))
+    return op.emitOpError("applicable operator is illegal");
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
 // CallOp
 //===----------------------------------------------------------------------===//
 
--- a/docs/EmitC.md
+++ b/docs/EmitC.md
@@ -18,6 +18,38 @@ Examples:
 
 ## Operation definition
 
+### `emitc.apply` (::mlir::emitc::ApplyOp)
+
+apply operation
+
+
+Syntax:
+
+```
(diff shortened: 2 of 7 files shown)


lib/Dialect/EmitC/IR/EmitCDialect.cpp
```diff
@@ -50,6 +50,18 @@ void emitc::buildTerminatedBody(OpBuilder &builder, Location loc) {
...
 
+//===----------------------------------------------------------------------===//
+// ApplyOp
+//===----------------------------------------------------------------------===//
+
+static LogicalResult verify(mlir::emitc::ApplyOp op) {
+  // Callee must not be empty.
+  if (op.applicableOperator().empty()) {
+    return op.emitOpError("applicable operator must not be empty");
+  }
```
simon-camp, line 61: Only allow "&" and "*" for now. We can then verify types between operand and result in a follow up (we might need a ptr type).

simon-camp (changes requested):
LGTM. We should just restrict the allowed operators I think.

lib/Dialect/EmitC/IR/EmitCDialect.cpp
marbre: Good point! Extended the verifier.

lib/Dialect/EmitC/IR/EmitCDialect.cpp
```diff
@@ -50,6 +50,24 @@ void emitc::buildTerminatedBody(OpBuilder &builder, Location loc) {
...
+
+  // Only `*` and `&` are supported.
+  StringRef applicableOperator = op.applicableOperator();
```
simon-camp, line 64: Move this to the top of the function and use it for the empty check as well.

lib/Dialect/EmitC/IR/EmitCDialect.cpp
```diff
@@ -50,6 +50,24 @@ void emitc::buildTerminatedBody(OpBuilder &builder, Location loc) {
   builder.create<emitc::YieldOp>(loc);
 }
 
+//===----------------------------------------------------------------------===//
+// ApplyOp
+//===----------------------------------------------------------------------===//
+
+static LogicalResult verify(mlir::emitc::ApplyOp op) {
+  // Callee must not be empty.
```
simon-camp, line 58: Change `callee` to `applicableOperator` or `operator`.

simon-camp (changes requested):

simon-camp:
In the PR description it should be `specify` instead of `specific`, ready to land after after these three nits :)

lib/Dialect/EmitC/IR/EmitCDialect.cpp
marbre: Good catch!

marbre:
> In the PR description it should be `specify` instead of `specific`, ready to land after after these three nits :)

All done. Thanks for going over with more care than me ;)

simon-camp (approved):

PR #136, opened 2021-05-07
Move dialect specific namespaces into `emitc`

marbre:
All our header only functions are actually reference implementations provided within the scope of `emitc`. Therefore, this moves the `mhlo`, `standard`, `tensor` and `tosa` namespaces into the top-level `emitc` namespace. The `emitc` namespace used in MIR-related parts will soon be moved to `mlir::emitc`.


dajuro (approved):
LGTM :+1:

PR #140, opened 2021-05-18
Add EmitC include operation

marbre:
This adds an include operation that allows to  define a source file
inclusion via `#include` directive. With the unit attribute
`is_standard_include` defined, the inclusion is emitted as
`#include <filename>`. Otherwise the inclusion is emitted as
`#include "filename"`.

### DIFF
--- a/test/Dialect/EmitC/ops.mlir
+++ b/test/Dialect/EmitC/ops.mlir
@@ -1,7 +1,7 @@
 // RUN: emitc-opt -allow-unregistered-dialect -verify-diagnostics %s | FileCheck %s
 
-"emitc.include" (){include = "<test.h>"} : () -> ()
-emitc.include "<test.h>"
+"emitc.include" (){include = "test.h", is_standard_include} : () -> ()
+emitc.include "test.h" is_standard_include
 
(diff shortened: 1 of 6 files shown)


test/Dialect/EmitC/ops.mlir
```diff
@@ -1,5 +1,8 @@
 // RUN: emitc-opt -allow-unregistered-dialect -verify-diagnostics %s | FileCheck %s
 
+"emitc.include" (){include = "<test.h>"} : () -> ()
+emitc.include "<test.h>"
```
simon-camp, line 4: nit: The angel brackets are irritating, as the type of include is handled by an attribute

```suggestion
"emitc.include" (){include = "test.h"} : () -> ()
emitc.include "test.h"
```

simon-camp (dismissed):
LGTM besides one nit :)

test/Dialect/EmitC/ops.mlir
marbre: Good catch! Missed to update this test while introducing the unit attribute. Fixed :)

simon-camp (approved):

PR #161, opened 2021-06-14
Address review comments

marbre:
Address (most) review comments of https://reviews.llvm.org/D103969.
Updating the code in our repo before updating (and splitting) the patch under review.

### DIFF
--- a/include/emitc/Dialect/EmitC/IR/EmitC.td
+++ b/include/emitc/Dialect/EmitC/IR/EmitC.td
@@ -30,9 +30,9 @@ class EmitC_Op<string mnemonic, list<OpTrait> traits = []> :
 }
 
 def EmitC_ApplyOp : EmitC_Op<"apply", []> {
-  let summary = "apply operation";
+  let summary = "Apply operation";
   let description = [{
-    With the "apply" operation the operators & (address of) and * (contents of)
+    With the `apply` operation the operators & (address of) and * (contents of)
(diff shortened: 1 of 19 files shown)


include/emitc/Dialect/EmitC/IR/EmitC.td
```diff
@@ -156,28 +62,25 @@ def EmitC_CallOp : EmitC_Op<"call", []> {
   let assemblyFormat = [{
     $callee `(` $operands `)` attr-dict `:` functional-type($operands, results)
   }];
-  let verifier = [{ return ::verify(*this); }];
 }
 
-def EmitC_ConstOp : EmitC_Op<"const", [ConstantLike, NoSideEffect]> {
-  let summary = "Constant op";
+def EmitC_ConstOp : EmitC_Op<"constant", [ConstantLike, NoSideEffect]> {
```
simon-camp, line 67:
```suggestion
def EmitC_ConstantOp : EmitC_Op<"constant", [ConstantLike, NoSideEffect]> {
```

simon-camp (changes requested):
For consistency I would also change the class name to `ConstantOp`.

marbre:
> For consistency I would also change the class name to `ConstantOp`.

I have to agree here. Thus, rebase and changed it accordingly.

simon-camp (approved):

PR #170, opened 2021-06-23
Address issues raised by clang-tidy

marbre:


include/emitc/Target/Cpp/CppEmitter.h
```diff
@@ -33,33 +33,32 @@ template <typename ForwardIterator, typename UnaryFunctor,
           typename NullaryFunctor>
 inline LogicalResult
 interleaveWithError(ForwardIterator begin, ForwardIterator end,
-                    UnaryFunctor each_fn, NullaryFunctor between_fn) {
+                    UnaryFunctor eachFn, NullaryFunctor betweenFn) {
```
simon-camp, line 36: This uses the exact same naming as an upstream function, see [here](https://llvm.org/doxygen/namespacellvm.html#ad75c79282c4734018fe1482b12d03d62).

lib/Dialect/EmitC/Conversion/MHLORegionOpsToEmitC.cpp
```diff
@@ -151,7 +151,7 @@ struct ConvertMhloRegionOpsToEmitCPass
     }
 
     OpBuilder builder(op);
-    auto ctx = op.getContext();
+    auto *ctx = op.getContext();
```
simon-camp, line 154: We will most likely remove auto anyway :)

simon-camp (changes requested):

include/emitc/Target/Cpp/CppEmitter.h
marbre: Indeed, but `clang-tidy` complains anyway and resulted in a build failure of https://reviews.llvm.org/D104632.

lib/Dialect/EmitC/Conversion/MHLORegionOpsToEmitC.cpp
marbre: Yes, it's just an intermediate change. I think it doesn't hurts.

include/emitc/Target/Cpp/CppEmitter.h
simon-camp: Okay that's a compelling reason.

simon-camp (approved):

PR #176, opened 2021-07-01
Integrate LLVM at llvm/llvm-project@e386871e

marbre:
Updates LLVM usage to match llvm/llvm-project@e386871e.
Further updates the MLIR-HLO submodule to tensorflow/mlir-hlo@a41d2374.


marbre:
Merging this requires to fix our header-only `tosa::matmul` implementation which needs support for 3d tensors.

PR #179, opened 2021-07-02
Rename `restrictedToC` and improve checks

marbre:
Renames `restrictedToC` to `isRestrictedToC`, adds additional checks to
prevent emitting `std::make_tuple` or `std::tie` if restricted to C.
Further extends the lit tests.


marbre:
We'll probably drop the mode restricting to C soon, but if not it's worth that this improvement isn't lost.

simon-camp (approved):

PR #180, opened 2021-07-02
Add batch dimension to tosa matmul implementation

simon-camp:


marbre (approved):
It would be nice to add additional unittests for a batch size other than 1. Can be done in a followup.

marbre (approved):

PR #200, opened 2021-08-24
Clarify which ops don't need a trailing semicolon

simon-camp:


marbre (changes requested):

marbre:
Sorry the linter trouble... I thought as it is a comment it would be acceptable -.-

marbre (approved):

PR #213, opened 2021-09-02
Integrate LLVM at llvm/llvm-project@2f0750d

marbre:
Updates LLVM usage to match llvm/llvm-project@2f0750d. This removes the
Cpp emitter which was upstreamed with commit llvm/llvm-project@2f0750d.


simon-camp (approved):
Nice

PR #228, opened 2021-11-15
Drop conversion for `arith.constant`

marbre:
As part of llvm/llvm-project@a54f4ea, native support to emit code for
`arith.constant` was added to the Cpp emitter.


dajuro (approved):
Looks good!

PR #229, opened 2021-11-30
Drop mhlo passes

marbre:
We only use the `mhlo-control-flow-to-scf` conversion pass from the
tensorflow/mlir-hlo repo. This drops all other mhlo passes.


marbre:
Seems we should look for an alternative in near future anyway, see https://github.com/google/iree/pull/7789.

simon-camp (approved):

PR #231, opened 2021-12-07
Add Eigen based implementation of tosa conv2d op

henri-gruender:
An optional replacement for the current implementation of the tosa conv2d op with better performance. Can be enabled using the CMake option `EMITC_TOSA_ENABLE_EIGEN`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ set(CMAKE_CXX_STANDARD 14)
 
 option(EMITC_BUILD_EMBEDDED "Build EmitC as part of another project" OFF)
 option(EMITC_ENABLE_HLO "Enables building MLIR-HLO." ON)
+option(EMITC_TOSA_ENABLE_EIGEN "Enables use of Eigen library for some TOSA Ops." OFF)
 option(EMITC_INCLUDE_TESTS "Generate build targets for the MLIR EmitC unit tests." ON)
 # TODO: Set to MLIR or LLVM default
 #       ${LLVM_INCLUDE_TESTS})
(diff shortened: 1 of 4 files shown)


marbre:
Our CI pipeline currently does not test this. At the moment, the CMake build tartget `MLIREmitCAllTests` is tested twice, when building in Debug and in Release mode. Furthermore, the model related test `Run TOSA test` which could be included for testing as well. So I actually dislike merging this without properly integrating it into our test strategy. Maybe we should split the `MLIREmitCAllTests` into different targets to at least tests it probably at this stage (must not be for Release and Debug?).

marbre (changes requested):
@henri-gruender Please see my [tosa-conv2d-eigen](https://github.com/marbre/mlir-emitc/tree/tosa-conv2d-eigen) branch or more specifically commit https://github.com/marbre/mlir-emitc/commit/f1aff589cd35302e29c6afbca22c12599a70abb1. I refactored your patch to allow a separate testing of the Eigen-based implementation without rerunning unrelated tests. However, on an Ubuntu 20.04 LTS with eigen installed via apt, it miserably fails to build.

henri-gruender:
That was a version issue with Eigen. The fixed version should work with Eigen >= 3.3.1

marbre (changes requested):
Please cherry-pick https://github.com/marbre/mlir-emitc/commit/048b64f44df9829cd19c5d5619efc9bcb08b592f. Otherwise we don't build and run your newly added implementation in our pipeline.

marbre:
@dajuro Please take an additional look.

dajuro (approved):
I just tested this: compiles without a warning and test succeeds. Nice work :+1:

marbre (approved):

PR #236, opened 2021-12-22
Add conversion for tosa arithmetic_right_shift

henri-gruender:


marbre (changes requested):
There should be no need to shift the `createBroadcastOpIfNeeded()` function. I suggest to rather move the the newly implemented function. The functions are not ordered by their function name anyway.

@dajuro Please review as well.

henri-gruender:
> There should be no need to shift the `createBroadcastOpIfNeeded()` function. I suggest to rather move the the newly implemented function. The functions are not ordered by their function name anyway.

Not by name, but I thought it would fit better under "common functions". Still I now undid the moving.

dajuro (approved):
Looks good to me :+1: I like how you untangled the logic of `arithmetic_right_shift`. It is way easier to read! Tests look good (I double checked them).

marbre:
> Looks good to me +1 I like how you untangled the logic of `arithmetic_right_shift`. It is way easier to read! Tests look good (I double checked them).

Thanks for reviewing! I will merge after the CI passed.

marbre:
> > There should be no need to shift the `createBroadcastOpIfNeeded()` function. I suggest to rather move the the newly implemented function. The functions are not ordered by their function name anyway.
> 
> Not by name, but I thought it would fit better under "common functions". Still I now undid the moving.

I agree, but this should be done in a separate commit. Otherwise it might be confusing as it is a change not related to what is described in the commit message.

marbre (approved):

PR #238, opened 2022-01-05
Add conversion for tosa logical_left_shift and clz

henri-gruender:


marbre (commented):
Even if these are element-wise ops, should we think about testing higher dimensional tensors? @dajuro @simon-camp What is your opinion?

marbre (approved):
We'll address the unit tests in a follow up.

PR #241, opened 2022-01-10
Build unit tests without `add_mlir_unittest`

marbre:
In preparation to move the EmitC header-only reference implementation
to its own CMake project, unit tests are now build with pure CMake.


henri-gruender (approved):
Works for me and looks good.

PR #245, opened 2022-01-14
Add conversion for tosa table

henri-gruender:


dajuro (approved):
Comparing this with the implementation in the TOSA reference model: this looks really good :+1:

PR #249, opened 2022-01-20
Update e2e_test.sh

heidebro:
Removes deprecated "--mhlo-control-flow-to-scf" commandline option from emitc-opt call.
Updates usage string with newest tested tensorflow version.


marbre (approved):
Good catch!

PR #251, opened 2022-01-20
Move the EmitC reference impl to its own folder

marbre:
This moves all `include/emitc/emitc_*.h` header files to
`reference-implementation/include/emitc/`. Furthermore, the `emitc_`
prefix on the header files is removed.


marbre:
I am also thing about moving the unit tests to the `reference-implementation` directory, but that is something for a followup.

dajuro (approved):

PR #254, opened 2022-01-25
Update tosa pad with explicit padding value

henri-gruender:
Adds the new optional parameter to the tosa pad operation, that was
added with tosa spec 0.23

Co-authored-by: Benjamin Heidebroek <benjamin.heidebroek@iml.fraunhofer.de>


marbre (commented):
I expected to see a change on the conversion side, but it is also nice to solve it with a default argument! Looks all good to me but I leave the final approval to @simon-camp to get an additional look, especially at the unit tests.

simon-camp (approved):
LGTM :+1:

PR #266, opened 2022-02-09
Clean up leftover of `convert-std-to-emitc`

marbre:
The `convert-std-to-emitc` pass was removed.


marbre:
Wow, that merge was by accident, but should be fine anyway.

PR #267, opened 2022-02-10
Refactor MHLO Unittests

heidebro:


reference-implementation/unittests/mhlo.cpp
```diff
@@ -31,160 +31,393 @@ const float EPSILON = 5e-4;
...
+    uint64_t expected_result = 1;
+    uint64_t result = mhlo::convert<uint64_t>(x);
 
-  Tensor0D<uint32_t> t0{1};
-  auto lambda_0d = [&t0]() -> Tensor0D<size_t> {
-    return mhlo::convert<Tensor0D<size_t>>(t0);
-  };
-  EXPECT_THAT(lambda_0d(), Pointwise(Eq(), {1}));
+    EXPECT_EQ(result, expected_result);
+  }
```
henri-gruender, line 120: This is the only test with blank lines between the scopes. Would be neater to remove them

reference-implementation/unittests/mhlo.cpp
```diff
@@ -31,160 +31,393 @@ const float EPSILON = 5e-4;
...
+    Tensor4D<double, 1, 2, 1, 2> expected_result{0.8775826f, 1.0f, 0.8775826f,
+                                                 1.0f};
+    Tensor4D<double, 1, 2, 1, 2> result = mhlo::cos(x);
+
+    EXPECT_THAT(result, Pointwise(FloatNear(EPSILON), expected_result));
+  }
 }
 
 TEST(mhlo, exponential) {
   EXPECT_NEAR(M_Ef32, mhlo::exponential(1.0f), EPSILON);
```
henri-gruender, line 209: Does not test negative values from here on

reference-implementation/unittests/mhlo.cpp
```diff
@@ -31,160 +31,393 @@ const float EPSILON = 5e-4;
...
-
-  EXPECT_THAT(mhlo::abs(t0), Pointwise(Eq(), {1}));
-  EXPECT_THAT(mhlo::abs(t1), Pointwise(Eq(), {1.0f, 2.0f}));
-  EXPECT_THAT(mhlo::abs(t2), Pointwise(Eq(), {2, 1, 0, 2}));
-  EXPECT_THAT(mhlo::abs(t3), Pointwise(Eq(), {2, 1, 0, 2}));
-  EXPECT_THAT(mhlo::abs(t4), Pointwise(Eq(), {2, 1, 0, 0, 3, 3, 2, 1}));
+  {
+    Tensor0D<int> x{-1};
+    Tensor0D<int> expected_result{1};
+    Tensor0D<int> result = mhlo::abs(x);
```
marbre, line 37: Should `Tensor0D` be the "lowest level" or should we also add tests for the fundamental datatype, similar to what we do in `arith.cpp`:
https://github.com/iml130/mlir-emitc/blob/d180553cc82b42f460ca8aea57491fc9d18c8fa2/reference-implementation/unittests/arith.cpp#L26-L39

simon-camp: We should check against the op [definitions](https://github.com/tensorflow/mlir-hlo/blob/402eeff6db40e3790eac6ff82b14fef67b86c0d7/include/mlir-hlo/Dialect/mhlo/IR/hlo_ops.td#L152) if scalar types are supported and change the [signature](https://github.com/iml130/mlir-emitc/blob/d180553cc82b42f460ca8aea57491fc9d18c8fa2/reference-implementation/include/emitc/mhlo.h#L603) if necessary.

In this case the abs function should be deleted for scalar types via a `IsTensor<Src> = true` template argument.

marbre: In that case, I would suggest to refactor our kernels in a follow up.

marbre: @heidebro Please fill an issue in our internal tracker.

henri-gruender (approved):

PR #271, opened 2022-02-24
Refactor tosa unittests

henri-gruender:
Part 1: Unary elementwise ops

### DIFF
--- a/reference-implementation/unittests/tosa.cpp
+++ b/reference-implementation/unittests/tosa.cpp
@@ -122,12 +122,52 @@ TEST(tosa, clz) {
 }
 
 TEST(tosa, reciprocal) {
-  Tensor<float, 4> t0{1.0f, 2.0f, 3.0f, 4.0f};
-  Tensor<float, 4> expected{1.0f, 0.5f, 0.3333f, 0.25f};
+  {
+    Tensor0D<float> x{1.0f};
+    Tensor0D<float> expected_result{1.0f};
+    Tensor0D<float> result = tosa::reciprocal(x);
 
-  Tensor<float, 4> result = tosa::reciprocal(t0);
+    EXPECT_THAT(result, Pointwise(FloatEq(), expected_result));
+  }
+  {
+    Tensor1D<double, 2> x{6.312247e+64, 9.053782e-32};
+    Tensor1D<double, 2> expected_result{1.5842219102009158e-65,
+                                        1.1045108000170537e+31};
+    Tensor1D<double, 2> result = tosa::reciprocal(x);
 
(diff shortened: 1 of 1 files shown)


reference-implementation/unittests/tosa.cpp
```diff
@@ -27,62 +28,192 @@ const float EPSILON = 5e-4;
...
-
-  Tensor<int64_t, 4, 2, 1> t1{-2, 2, -2, 3, 4, -5, 5, 5};
-  int64_t min1 = 1;
-  int64_t max1 = 3;
-  Tensor<int64_t, 4, 2, 1> expected_result1{1, 2, 1, 3, 3, 1, 3, 3};
-  Tensor<int64_t, 4, 2, 1> s1 = tosa::clamp(t1, min1, max1);
-  EXPECT_THAT(s1, Pointwise(Eq(), expected_result1));
+  {
+    Tensor0D<int32_t> operand{-1};
+    int32_t min_value = -3;
```
heidebro, line 33: Should we generally define for our internal style guide that parameter variables need to be defined in the same order that they appear in in the function signature?
Otherwise reordering might be nicer visually, but that's a nitpick.

reference-implementation/unittests/tosa.cpp
```diff
@@ -27,62 +28,192 @@ const float EPSILON = 5e-4;
...
+
+    EXPECT_THAT(result, Pointwise(Eq(), expected_result));
+  }
 }
 
 TEST(tosa, reciprocal) {
-  Tensor<float, 4> t0{1.0f, 2.0f, 3.0f, 4.0f};
-  Tensor<float, 4> expected{1.0f, 0.5f, 0.3333f, 0.25f};
+  {
+    Tensor0D<float> x{1.0f};
```
heidebro, line 126: Shouldn't negative values be tested too for this function?

reference-implementation/unittests/tosa.cpp
henri-gruender: I think having the arguments in the same order as in the signature is a good rule. Which way it looks nicer is quite subjective.

reference-implementation/unittests/tosa.cpp
henri-gruender: Yes!

reference-implementation/unittests/tosa.cpp
marbre: Let's order the arguments according to the function signature :)

heidebro (approved):
Looks good now! LGTM

marbre (approved):

PR #275, opened 2022-03-08
Integrate LLVM at llvm/llvm-project@1a42383

marbre:
Updates LLVM usage to match llvm/llvm-project@1a42383.
Further updates the MLIR-HLO submodule to tensorflow/mlir-hlo@513b053.


docs/func-op-coverage.md
```diff
@@ -1,11 +1,9 @@
...
 
-The table below shows the supported Standard ops.
+The table below shows the supported func ops.
 
 | op                    | supported          | comment |
 | :-------------------- |:------------------:| :------ |
-| br                    | :white_check_mark: | via `emitc-translate` |
 | call                  | :white_check_mark: | via `emitc-translate` |
-| cond_br               | :white_check_mark: | via `emitc-translate` |
 | constant              | :white_check_mark: | via `emitc-translate` |
```
simon-camp, line 8: I don't think we support constants for [function symbols](https://mlir.llvm.org/docs/Dialects/Func/#funcconstant-mlirfuncconstantop). Seems to be an overlook from the arith Dialect split.

```suggestion
```

simon-camp (changes requested):

docs/func-op-coverage.md
marbre: We actually do support `func::ConstantOp`, see https://github.com/llvm/llvm-project/blob/406d418c0c831287b41c83150369fd8485c3d700/mlir/lib/Target/Cpp/TranslateToCpp.cpp#L924 and https://github.com/llvm/llvm-project/blob/406d418c0c831287b41c83150369fd8485c3d700/mlir/lib/Target/Cpp/TranslateToCpp.cpp#L240-L246. It might be that there is no full support, but this what came with the dialect renaming llvm/llvm-project@23aa5a7.

PR #281, opened 2022-03-21
Expand `mhlo::select` and refactor unit tests

### LINKED ISSUE #280: MHLO select scalar pred support
select allows for scalar predicate support while reference implementation does not.

marbre:
According to
https://www.tensorflow.org/mlir/hlo_ops#mhloselect_mlirmhloselectop the
tensor of pred may be a scalar in which case it is broadcasted. Instead
of broadcasting we keep it a 0D tensor and overload the reference
implementation with a function that handles it accordingly.

Fixes #280


simon-camp (approved):
LGTM :+1:

jpienaar:
Thanks

PR #283, opened 2022-03-22
Fix building with `EMITC_ENABLE_HLO` set to OFF

marbre:


marbre:
It is required to have a function declaration in the header. The definition isn't needed as we don't link it into the `emitc-opt`/`emitc-translate` tools when `EMITC_ENABLE_HLO` is set to OFF.

simon-camp (approved):

PR #285, opened 2022-03-24
Add support for mhlo reduce op with two results in reference implementation

simon-camp:


dajuro (approved):
LGTM :+1:

marbre:
Thanks for reviewing @dajuro :)

Issue #279, opened 2022-03-16
Add support for `tosa.argmax`


marbre:
There is currently no support for [`tosa.argmax`](https://mlir.llvm.org/docs/Dialects/TOSA/#tosaargmax-mlirtosaargmaxop). Support needs to be added to the reference implementation ([reference-implementation/include/emitc/tosa.h](https://github.com/iml130/mlir-emitc/blob/main/reference-implementation/include/emitc/tosa.h)) and a conversion pattern needs to added.

marbre:
Closed with #291.

PR #291, opened 2022-04-27
Add support for tosa argmax operation

simon-camp:
Fixes #291.

### DIFF
--- a/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
+++ b/lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
@@ -607,8 +607,9 @@ class ReduceOpConversion : public OpConversionPattern<SrcOp> {
   using OpConversionPattern<SrcOp>::OpConversionPattern;
 
 public:
-  ReduceOpConversion(MLIRContext *ctx, StringRef funcName)
-      : OpConversionPattern<SrcOp>(ctx), funcName(funcName) {}
+  ReduceOpConversion(MLIRContext *ctx, StringRef funcName, bool keepDims)
+      : OpConversionPattern<SrcOp>(ctx), funcName(funcName),
+        keepDims(keepDims) {}
 
 private:
   LogicalResult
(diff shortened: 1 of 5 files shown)


test/Conversion/tosa-to-emitc.mlir
```diff
@@ -238,6 +238,12 @@ func @test_matmul(%arg0: tensor<1x14x19xf32>, %arg1: tensor<1x19x28xf32>) -> ten
 }
 
 // Reduce ops
+func @test_argmax(%arg0: tensor<13x21x3xi32>) -> tensor<21x3xi32> {
```
marbre, line 241: NIT: This isn't a reduce op, thus the test should be inserted before the comment above (or the comment should be moved).

marbre (changes requested):

lib/Dialect/EmitC/Conversion/TosaToEmitC.cpp
```diff
@@ -606,8 +606,9 @@ class ReduceOpConversion : public OpConversionPattern<SrcOp> {
   using OpConversionPattern<SrcOp>::OpConversionPattern;
 
 public:
-  ReduceOpConversion(MLIRContext *ctx, StringRef funcName)
-      : OpConversionPattern<SrcOp>(ctx), funcName(funcName) {}
+  ReduceOpConversion(MLIRContext *ctx, StringRef funcName, bool keepDims)
```
dajuro, line 609: Is a reduction without dropping reduced dimensions still a reduction?

test/Conversion/tosa-to-emitc.mlir
marbre: To be more precise, `// Reduce ops` here only refers to ops that have `reuduce` in their name.

marbre:
@dajuro and @heidebro can one of you explicitly look at the function? I only had a look to the mechanics.

@simon-camp please also add a note to https://github.com/iml130/mlir-emitc/blob/main/docs/tosa-op-coverage.md :)

dajuro (approved):

marbre:
> Fixes #291.

Should have been #279. #291 links to this merge request :wink:

PR #300, opened 2022-06-03
Add conversion for tosa rescale

henri-gruender:


reference-implementation/include/emitc/tosa.h
```diff
@@ -114,6 +114,46 @@ inline Src reluN(Src operand, typename Src::value_type max_value) {
...
+// RescaleOp
+template <typename Dest, size_t Dim, typename Src>
+inline Dest rescale(Src x, typename get_element_type<Src>::type in_zp,
+                    typename get_element_type<Dest>::type out_zp,
+                    Tensor1D<int64_t, Dim> mult, Tensor1D<int64_t, Dim> shift,
+                    bool scale32, bool double_round, bool per_channel) {
+  using ET_Dest = typename get_element_type<Dest>::type;
+  using Dest_I32 = typename replace_element_type<int32_t, Dest>::type;
+
+  double_round &= scale32;
```
simon-camp, line 126: Could you replace this with a runtime assertion?

The [docs](https://www.mlplatform.org/tosa/tosa_spec.html#_rescale) state that `!scale32 && double_round` should result in an error.

```suggestion
  if(!scale32 && double_round) {
    assert(false && "Invalid combination of `scale32` and `double_round` arguments.");  
  }
```

That makes it easier to relate the code to the docs.

reference-implementation/include/emitc/tosa.h
```diff
@@ -114,6 +114,46 @@ inline Src reluN(Src operand, typename Src::value_type max_value) {
...
+    int64_t result = (element * mult + round) >> shift;
+    return static_cast<int32_t>(result);
+  };
+
+  Dest_I32 result;
+  for (size_t i = 0; i < x.size(); ++i) {
+    size_t index = per_channel ? x.unravel_index(i)[x.rank() - 1] : 0;
+    int64_t element = x[i] - in_zp;
+    result[i] = apply_scale(element, mult[index], shift[index]);
+    result[i] += out_zp;
```
simon-camp, line 146: Maybe something like the following looks a bit better instead of writing the result array twice.

```suggestion
    int32_t scaled_element = apply_scale(element, mult[index], shift[index]);
    result[i] = scaled_element + out_zp;
```

simon-camp, line 146: Maybe something like the following looks a bit better instead of writing the result array twice.

```suggestion
    int32_t scaled_element = apply_scale(element, mult[index], shift[index]);
    result[i] = scaled_element + out_zp;
```

simon-camp (changes requested):
This looks good, just two minor suggestions.

reference-implementation/include/emitc/tosa.h
henri-gruender: Yes, but why not put the error condition inside the assertion? Should I check for the other error cases from the docs, too?

henri-gruender: We decided that only checking this case will be the most consistent.

marbre (approved):

simon-camp (approved):

PR #306, opened 2022-06-24
Cleanup test config

marbre:


simon-camp (approved):
:+1: I've tested this in an llvm external build.

PR #315, opened 2022-07-25
[nfc] Flip accessors to prefixed form

marbre:
This flips the accessors used to the prefixed form, similar as upstream
recently did.


marbre:
Depends on #314.

simon-camp (approved):
LGTM once compilation passes.

PR #324, opened 2022-08-22
Remove ReluN conversion and implementation

marbre:
The ReluN operator was removed from the TOSA spec and the MLIR op is
removed with llvm/llvm-project@9dec80b. This patch therefore drops the
conversion to EmitC as well as the reference implementation. The latter
already relied on clamp. Hence, tests are moved instead of dropped.

### DIFF
--- a/reference-implementation/unittests/tosa.cpp
+++ b/reference-implementation/unittests/tosa.cpp
@@ -80,6 +80,60 @@ TEST(tosa, clamp) {
 
     EXPECT_THAT(result, Pointwise(FloatEq(), expected_result));
   }
+
+  // These are the tests of the former ReluN operation, which has been removed
+  // from the TOSA specification. clamp with min_value=0 can be used instead.
+  {
+    Tensor0D<int32_t> operand{0};
+    int32_t max_value = 0;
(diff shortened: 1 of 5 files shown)


dajuro (approved):
Beautiful :+1:

reference-implementation/unittests/tosa.cpp
```diff
@@ -80,6 +80,59 @@ TEST(tosa, clamp) {
...
+  // These are the tests of the former ReluN operation, which has been removed
+  // from the TOSA specification. clamp with min_value=0 can be used instead.
+  {
+    Tensor0D<int32_t> operand{0};
+    int32_t max_value = 0;
+    Tensor0D<int32_t> expected_result{0};
+    Tensor0D<int32_t> result = tosa::clamp(operand, {0}, max_value);
+
+    EXPECT_THAT(result, Pointwise(Eq(), expected_result));
+  }  {
```
dajuro, line 93: This needs to be formatted!

marbre: Yep.. the linter action was broken and thus I didn't noticed before. This is fixed now (with the linter action revised in the background).

PR #328, opened 2022-08-24
Enable MHLO in LLVM external project builds

marbre:
Allows to set `EMITC_ENABLE_HLO` when using the `LLVM_EXTERNAL_PROJECTS`
mechanism.


simon-camp (approved):
I've built and run the MHLO tests locally as external project.

PR #330, opened 2022-08-29
Fix build warnings

simon-camp:
Eliminates warnings previously reported with `-Wbraced-scalar-init`.


marbre (approved):
I was just about to push the same fix :laughing:

PR #336, opened 2022-11-17
Add conversion for tosa equal

NatashaKnk:


marbre (changes requested):
Thanks for your interest in the project and contributing to it @NatashaKnk! I am oof today and will review more carefully early next week.

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -768,6 +768,8 @@ void populateTosaToEmitcPatterns(MLIRContext *ctx,
                                                            "emitc::tosa::add");
   patterns.add<ArithmeticRightShiftOpConversion>(
       ctx, "emitc::tosa::arithmetic_right_shift");
+  patterns.add<CallOpBroadcastableConversion<tosa::EqualOp>>(
+      ctx, "emitc::tosa::equal");
```
marbre, line 772: With the current patch
```mlir
%0 = "tosa.equal"(%arg0, %arg1) : (tensor<13x21x3xi32>, tensor<13x21x3xi32>) -> tensor<13x21x3xi1>
```
is translated into

```C++
  Tensor<bool, 13, 21, 3> v3 = emitc::tosa::equal(v1, v2);
```
Looking into the header-only implementation you're providing with this PR, the return type needs to be passed as a template argument. This is currently not the case. However, the `CallOpConversion` as well as the `CallOpBroadcastableConversion` can handle this. With
```suggestion
      ctx, "emitc::tosa::equal", /*explicitResultType=*/true);
```
the mlir snippet is translated to:
```
Tensor<bool, 13, 21, 3> v3 = emitc::tosa::equal<Tensor<bool, 13, 21, 3>>(v1, v2);
```

test/Conversion/tosa-to-emitc.mlir
```diff
@@ -125,6 +125,12 @@ func.func @test_arithmetic_right_shift2(%arg0: tensor<13x21x1xi32>, %arg1: tenso
   return %0 : tensor<13x21x3xi32>
 }
 
+func.func @test_equal(%arg0: tensor<13x21x3xi32>, %arg1: tensor<13x21x3xi32>) -> tensor<13x21x3xi1> {
+  // CHECK: emitc.call "emitc::tosa::equal"(%arg0, %arg1) : (tensor<13x21x3xi32>, tensor<13x21x3xi32>) -> tensor<13x21x3xi1>
```
marbre, line 129: With the code change in the conversion, the test must be adjusted to the following:

```suggestion
  // CHECK: emitc.call "emitc::tosa::equal"(%arg0, %arg1) {template_args = [tensor<13x21x3xi1>]} : (tensor<13x21x3xi32>, tensor<13x21x3xi32>) -> tensor<13x21x3xi1>
```

marbre (changes requested):
Thanks again for your patch. Having time to review it more carefully it turns out that there is still a fix required. Please see my suggestions. The pipeline passed before as there is no end-to-end test for the `tosa.equal` op (yet).
For testing end-to-end, we're using MobileNetV2 which doesn't contain a `tosa.equal` op. So no need to add an end-to-end test as part if this pull request. However, I would highly appreciate if we could jointly work on improving our end-to-end testing as soon as you add further ops.

marbre (approved):
Looks good, thanks for your patch @NatashaKnk!

PR #338, opened 2022-12-02
Add lowering for tosa.tile

NatashaKnk:
This implementation uses a not-so-pretty amount of for-loops, especially for the 4D case. The alternative option is iteration over the output in a single for-loop, which requires some more complex computation in order to reference the input index to read from. The current version seems more readable, however if there are benefits to the alternative that I'm missing I'd gladly swap over to it.


lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -739,6 +739,36 @@ class SliceOpConversion : public OpConversionPattern<tosa::SliceOp> {
...
+private:
+  LogicalResult
+  matchAndRewrite(tosa::TileOp tileOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    StringAttr callee = rewriter.getStringAttr("emitc::tosa::tile");
+
+    // clang-format off
+    ArrayAttr args = rewriter.getArrayAttr({
+      rewriter.getIndexAttr(0),
+      getI64ElementsAttr(tileOp.getMultiplesAttr(), tileOp.getContext()),
```
marbre, line 758:
```suggestion
      getI32ElementsAttr(tileOp.getMultiplesAttr(), tileOp.getContext()),
```
The [TOSA 0.40.0 specification](https://www.mlplatform.org/tosa/tosa_spec.html#_tile) states that the multiplies attribute should be of type `int32_t`. I've just pushed a `getI32ElementsAttr` helper that can be used here.

test/Conversion/tosa-to-emitc.mlir
```diff
@@ -325,6 +325,12 @@ func.func @test_reshape(%arg0: tensor<13x21x3xf32>) -> tensor<1x819xf32> {
   return %0 : tensor<1x819xf32>
 }
 
+func.func @test_tile(%arg0: tensor<1x3x1x4xf32>) -> tensor<2x3x3x8xf32> {
+  // CHECK: emitc.call "emitc::tosa::tile"(%arg0) {args = [0 : index, dense<[2, 1, 3, 2]> : tensor<4xi64>], template_args = [tensor<2x3x3x8xf32>]} : (tensor<1x3x1x4xf32>) -> tensor<2x3x3x8xf32>
```
marbre, line 329:
```suggestion
  // CHECK: emitc.call "emitc::tosa::tile"(%arg0) {args = [0 : index, dense<[2, 1, 3, 2]> : tensor<4xi32>], template_args = [tensor<2x3x3x8xf32>]} : (tensor<1x3x1x4xf32>) -> tensor<2x3x3x8xf32>
```

marbre (changes requested):
I agree that the for-loops are not so pretty but I am fine with preferring readability here. The implementation looks good to me in general. However, I couldn't had the time to review all unittests but I asked @henri-gruender to support with this, especially for the 3D and 4D cases.

NIT: According to the TOSA specification, the type of the `multiplies` attribute should be an `int32_t`, see my inline comments.

henri-gruender (approved):
I didn't find anything wrong with the unit tests.

marbre:
> I didn't find anything wrong with the unit tests.

Thanks for helping out with the review!

NatashaKnk:
Whoops, I entirely missed that! Thanks for pointing it out and adding support for it :)

marbre (changes requested):
Thanks for submitting the revised patch! I think with the code generation changed, the reference implementation as well as the unit tests need to be updated as well. Without having tested it, I think the `multiplies` attribute/parameter should be of type `int32_t`.
I haven't checked the type of `Src:dim()` and with the suggested changes `int32_t` and `int64_t` are used in the index calculation. Thus I assume that further changes are necessary.

NatashaKnk:
Src::dim() returns a size_t. AFAIK that defaults to i64, but I'm not entirely sure whether size_t is supposed to handle this kind of situation. Do you believe that creating a template for shape() that supports explicit return types would be a good approach? (as casting each dim, especially in a for loop doesn't seem like a very elegant solution)

marbre:
> Src::dim() returns a size_t. AFAIK that defaults to i64, but I'm not entirely sure whether size_t is supposed to handle this kind of situation. Do you believe that creating a template for shape() that supports explicit return types would be a good approach? (as casting each dim, especially in a for loop doesn't seem like a very elegant solution)

You’re right, `dim()` returns a `size_t`
https://github.com/iml130/mlir-emitc/blob/83cc8462882b64ae600362be23f7cd7bd316fd9b/reference-implementation/include/emitc/types.h#L113
and `size_t` _normally_ defaults to an unsigned integer (`unsigned int` or even `unsigned long int`). In general, adding an unsigned integer to a `int32_t` is of course not a good thing, but I don’t know how likely it is that we will ever hit this…

Regarding a template for shape(), I agree that casting each dim isn’t elegant. However, I don’t had a plan in mind when I started reviewing your patch. I just tripped over the return type of multiplies when looking into the TOSA spec. Anyway, we don’t to be strictly compliant to the spec. If it makes the life easier for us we can also adjust the type of `multiplies` to whatever works best for us (in addition we should state somewhere that we’re not 100% complaint).
TL;DR whatever works best for you as you’re the only user of `tosa::tile` at the moment.

jpienaar:
Dimensions exceeding 32-bit is rare, so indeed this is possible. I'd suggest that if we follow https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop we end up hoisting and casting some of these outside the loops and if we are really concerned we could add asserts to verify. Alternatively, we could verify the shape of the ops we lower from, so we only lower to these in case the shape can be addressed with 32-bit dims and so avoid generating the emitc one where this could fail.

NatashaKnk:
Left the casting to be implicit, added a check in TosaToEmitC to test that none of the dimensions are larger than the i32 limit. Hope this seems reasonable. Thanks!

marbre (approved):
> Left the casting to be implicit, added a check in TosaToEmitC to test that none of the dimensions are larger than the i32 limit. Hope this seems reasonable. Thanks!

As decided in our off GitHub discussion, we'll try to stick to the TOSA specs and **not** to the TOSA dialect in MLIR core. Apart from this, we could implement a verifier pass instead of checking tensor dimensions within the conversion pass, but this is fine for now.

PR #340, opened 2022-12-16
Implement tosa lowerings for gather, select and concat ops.

NatashaKnk:
* As MHLO already had an implementation for concatenate and the inputs/outputs are identical, moved the existing MHLO concatenate implementation into core_ops and referenced it from both. 
* The createBroadcastOpIfNeeded was initially implemented to support binary ops only. However, very small changes were required to enable it in supporting broadcasting for more than two inputs. This can be further expanded by allowing callOpBroadcastableConversion to take in more than two inputs, but it may require more substantial changes that might not be worth it for (what is currently) a single op.

### DIFF
--- a/docs/tosa-op-coverage.md
+++ b/docs/tosa-op-coverage.md
@@ -38,8 +38,8 @@ The table below shows the supported TOSA ops.
 | concat                 | :heavy_check_mark: | |
 | conv2d                 | :white_check_mark: | Quantization and dilation not supported |
 | depthwise_conv2d       | :white_check_mark: | Quantization and dilation not supported |
-| gather                 | :heavy_check_mark: | |
 | fully_connected        | :white_check_mark: | Quantization not supported |
+| gather                 | :heavy_check_mark: | |
 | matmul                 | :white_check_mark: | Quantization not supported |
 | reduce_all             | :heavy_check_mark: | |
 | reduce_any             | :heavy_check_mark: | |
--- a/docs/tosa-op-coverage.md
+++ b/docs/tosa-op-coverage.md
@@ -31,10 +31,14 @@ The table below shows the supported TOSA ops.
 | pow                    | :heavy_check_mark: | |
 | sub                    | :heavy_check_mark: | |
 | table                  | :heavy_check_mark: | |
+| **Ternary elementwise ops**
+| argmax                 | :heavy_check_mark: | |
 | **Other ops**
 | argmax                 | :heavy_check_mark: | |
+| concat                 | :heavy_check_mark: | |
 | conv2d                 | :white_check_mark: | Quantization and dilation not supported |
 | depthwise_conv2d       | :white_check_mark: | Quantization and dilation not supported |
+| gather                 | :heavy_check_mark: | |
 | fully_connected        | :white_check_mark: | Quantization not supported |
 | matmul                 | :white_check_mark: | Quantization not supported |
 | reduce_all             | :heavy_check_mark: | |
--- a/docs/tosa-op-coverage.md
+++ b/docs/tosa-op-coverage.md
@@ -32,7 +32,7 @@ The table below shows the supported TOSA ops.
 | sub                    | :heavy_check_mark: | |
 | table                  | :heavy_check_mark: | |
 | **Ternary elementwise ops**
-| argmax                 | :heavy_check_mark: | |
+| select                 | :heavy_check_mark: | |
 | **Other ops**
 | argmax                 | :heavy_check_mark: | |
 | concat                 | :heavy_check_mark: | |
--- a/lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
+++ b/lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
@@ -134,7 +134,7 @@ private:
 
     ArrayAttr args;
     ArrayAttr templateArgs =
-        rewriter.getArrayAttr({rewriter.getI64IntegerAttr(concatOp.getAxis()),
+        rewriter.getArrayAttr({concatOp.getAxisAttr(),
                                TypeAttr::get(concatOp.getResult().getType())});
 
     rewriter.replaceOpWithNewOp<emitc::CallOp>(concatOp, concatOp.getType(),
@@ -441,10 +441,9 @@ private:
   }
 };
 
-/// Convert `tosa.fully_connected` into an `emitc.call` operation.
-template <size_t operands, typename SrcOp,
-          typename Adaptor = typename SrcOp::Adaptor>
-SmallVector<Value, operands>
+/// Creates a broadcast operation if the inputs don't match the output shape.
+template <typename SrcOp, typename Adaptor = typename SrcOp::Adaptor>
+SmallVector<Value>
 createBroadcastOpIfNeeded(SrcOp &srcOp, Adaptor adaptor,
                           ConversionPatternRewriter &rewriter) {
   // TOSA allows implicit broadcasting, so we need to insert broadcast_in_dim
@@ -459,7 +458,8 @@ createBroadcastOpIfNeeded(SrcOp &srcOp, Adaptor adaptor,
   Value output = srcOp.getResult();
   auto opOutputShape = output.getType().cast<RankedTensorType>().getShape();
   auto opOutputRank = output.getType().cast<RankedTensorType>().getRank();
-  SmallVector<Value, operands> broadcastedOperands;
+  size_t operands = srcOp.getNumOperands();
+  SmallVector<Value> broadcastedOperands;
 
   for (auto operand : adaptor.getOperands()) {
     RankedTensorType operandTensor =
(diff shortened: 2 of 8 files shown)


docs/tosa-op-coverage.md
```diff
@@ -31,10 +31,14 @@ The table below shows the supported TOSA ops.
...
 | sub                    | :heavy_check_mark: | |
 | table                  | :heavy_check_mark: | |
+| **Ternary elementwise ops**
+| select                 | :heavy_check_mark: | |
 | **Other ops**
 | argmax                 | :heavy_check_mark: | |
+| concat                 | :heavy_check_mark: | |
 | conv2d                 | :white_check_mark: | Quantization and dilation not supported |
 | depthwise_conv2d       | :white_check_mark: | Quantization and dilation not supported |
+| gather                 | :heavy_check_mark: | |
```
jpienaar, line 41: Nit: gather below fully_connected to keep these in order.

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -118,6 +118,33 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+  LogicalResult
+  matchAndRewrite(tosa::ConcatOp concatOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    StringRef funcName = "emitc::tosa::concat";
+    StringAttr callee = rewriter.getStringAttr(funcName);
+
+    ArrayAttr args;
+    ArrayAttr templateArgs =
+        rewriter.getArrayAttr({rewriter.getI64IntegerAttr(concatOp.getAxis()),
```
jpienaar, line 137: concatOp.getAxisAttr() should already be a I64Attr

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -415,8 +442,9 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
 };
 
 /// Convert `tosa.fully_connected` into an `emitc.call` operation.
```
jpienaar, line 444: This comment seems off.

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -415,8 +442,9 @@ class RsqrtOpConversion : public OpConversionPattern<tosa::RsqrtOp> {
...
-template <typename SrcOp, typename Adaptor = typename SrcOp::Adaptor>
-SmallVector<Value, 2>
+template <size_t operands, typename SrcOp,
```
jpienaar, line 445: Why do we need to pass operands here in template? SmallVector can be resized and seems like it has to match Adaptor::getNumOperands().

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -118,6 +118,33 @@ class CallOpConversion : public OpConversionPattern<SrcOp> {
...
+
+public:
+  ConcatOpConversion(MLIRContext *ctx) : OpConversionPattern(ctx) {}
+
+private:
+  LogicalResult
+  matchAndRewrite(tosa::ConcatOp concatOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    StringRef funcName = "emitc::tosa::concat";
```
jpienaar, line 132: Consider using StringLiteral.

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -610,6 +641,36 @@ class ArithmeticRightShiftOpConversion
...
+
+    ArrayAttr args;
+    ArrayAttr templateArgs;
+
+    SmallVector<Value, 3> broadcastedOperands =
+        createBroadcastOpIfNeeded<3>(selectOp, adaptor, rewriter);
+
+    rewriter.replaceOpWithNewOp<emitc::CallOp>(
+        selectOp, selectOp.getType(), callee, args, templateArgs,
+        ValueRange({broadcastedOperands[0], broadcastedOperands[1],
```
jpienaar, line 667: Aren't these all of broadcastedOperands? I think a ValueRange should automatically be constructed here.

reference-implementation/include/emitc/types.h
```diff
@@ -292,6 +292,34 @@ inline Dest binary(const SrcLeft &x, const SrcRight &y, BinaryOp &&op) {
...
+inline Dest ternary(const SrcA &a, const SrcB &b, const SrcB &c,
+                    TernaryOp &&op) {
+  Dest d;
+  auto first1 = a.begin(), last1 = a.end();
+  auto first2 = b.begin(), first3 = c.begin();
+  auto result = d.begin();
+
+  while (first1 != last1) {
+    *result = op(*first1, *first2, *first3);
+    result++;
```
jpienaar, line 315: Nit: prefer pre-increment unless post-increment behavior is needed.

(almost feels like this could have simpler with llvm::zip to bring all together, especially as you probably need to assert that we haven't hit the end of any of these).

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
marbre: Since we use `StringRef funcName` at many places in our code you could leave as is. I will refactor the code anyway ad using `StringLiteral` is the better way.

marbre:
As I am already oof this year and thus have only limited time to review, @henri-gruender will assist with reviewing the reference implementation and its unit tests.

henri-gruender (changes requested):
Looks good in general.
For the unit tests we try to have test cases for the possible ranks of the tensors and with float and integer types. I think these tests are still fine for now. `concat` already has good tests from the mhlo implementation.

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
NatashaKnk: Oops, I actually fixed it before but it got lost somewhere in the process. Fixed it again :)

lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
NatashaKnk: Good point, for some reason I had assumed we would need to specify the SmallVector size. Fixed, thanks!

reference-implementation/include/emitc/types.h
NatashaKnk: I was thinking the same thing, but I wasn't entirely sure how to wire llvm::zip into here.

reference-implementation/include/emitc/tosa.h
```diff
@@ -270,7 +270,30 @@ inline Tensor<int32_t, Shape...> table(Tensor<int16_t, Shape...> x,
...
+template <int32_t Dimension, typename Dest, typename... Src>
+inline Dest concat(Src... inputs) {
+  return emitc::concatenate<Dimension, Dest, Src...>(inputs...);
+}
+
+template <int32_t Dimension, typename Dest, typename Src1, typename... Src>
+inline Dest concat(Src1 input1, Src... inputs) {
+  return emitc::concatenate<Dimension, Dest, Src1, Src...>(input1, inputs...);
+}
+
```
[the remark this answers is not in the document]
NatashaKnk: Good point! Thanks.

reference-implementation/include/emitc/tosa.h
```diff
@@ -475,6 +498,33 @@ Dest fully_connected(Src input, Weights weights, Bias bias) {
...
+  static_assert(input.dim(0) == indices.dim(0),
+                "Input and weight batch dimension do not match.");
+  static_assert(input.dim(2) == result.dim(2),
+                "Input and output channel dimension do not match.");
+  static_assert(indices.dim(1) == result.dim(1),
+                "Weight and output index dimension do not match.");
+
+  auto it = result.begin();
+  size_t d0offset = Src::dim(1) * Src::dim(2);
+  for (int64_t i = 0, idx = Idx::size(); i < idx; i++) {
```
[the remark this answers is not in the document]
NatashaKnk: Fixed, thanks!

NatashaKnk:
Added a couple of tests, but not entirely sure what other ones should be added. Since a lot of the ops operate on Tosa_Tensor which doesn't have a specific range of possible ranks (as far as I'm aware), how many potential ranks would be considered sufficient? Similarly, when it comes to types, should I test for each potential bit width, or is an all-encompassing int/float test enough?
Any other testing tips are also appreciated :)

henri-gruender:
> Since a lot of the ops operate on Tosa_Tensor which doesn't have a specific range of possible ranks (as far as I'm aware), how many potential ranks would be considered sufficient?

Right, most common are probably Tensor0D through Tensor4D (which are also the ones defined by name in the Tensor class). So if there is no info about the possible ranks, you could stick to those.

> Similarly, when it comes to types, should I test for each potential bit width, or is an all-encompassing int/float test enough?

In my opinion it is enough, if there is at least one test case for an int of any width and one for float or double.

Overall I think the first five test cases for `tosa::clamp` set a good example for a general function, but it is more important to look for edge cases in the implementation and specification. Hopefully we can add some guidelines for the tests in the near future, but for now the current set of tests in the pull request is good enough as it is.

henri-gruender (approved):

marbre (approved):

PR #341, opened 2023-01-04
Add tosa.greater_equal lowering to EmitC

NatashaKnk:


lib/Conversion/TosaToEmitC/TosaToEmitC.cpp
```diff
@@ -881,6 +881,8 @@ void populateTosaToEmitcPatterns(MLIRContext *ctx,
                                                            "emitc::tosa::add");
   patterns.add<ArithmeticRightShiftOpConversion>(
       ctx, "emitc::tosa::arithmetic_right_shift");
+  patterns.add<CallOpBroadcastableConversion<tosa::GreaterEqualOp>>(
```
jpienaar, line 884: Can we have this below the Equal case? (Keeping these a bit more sorted will make it easier).

jpienaar (approved):
Looks good thanks!

marbre (approved):
Thanks! LGTM as well. I only had a minor remark but changed this straight away instead of suggesting changes. Hope this is okay for you.

PR #343, opened 2023-01-11
Integrate LLVM at llvm/llvm-project@8bd65e5

marbre:
Updates LLVM usage to match llvm/llvm-project@8bd65e5.
* TOSA ops switched to DenseArrayAttr with commit llvm/llvm-project@9e1a344. The types of attributes are therfore aligned in reference implementation.
* Uses std::nullopt instead of None.
* Moves from llvm::makeArrayRef to ArrayRef deduction guides, see https://reviews.llvm.org/D140896.

Further updates the MLIR-HLO submodule to tensorflow/mlir-hlo@78c6dcf.


marbre:
@NatashaKnk this would change reference implementation for the tosa.tile op as `tileOp.getMultiplesAttr()` now returns an DenseArrayAttr which is emitted as `int64_t` by EmitC. Would this be okay for you or do we need to handle this in a different way? I am aware that the spec still states it should be an `int32_t`...

reference-implementation/unittests/tosa.cpp
```diff
@@ -278,13 +278,17 @@ TEST(tosa, rescale) {
 
     EXPECT_THAT(result, Pointwise(Eq(), expected_result));
   }
+  // TODO: Fix test
```
simon-camp, line 281:
```suggestion
  // TODO(#344): Fix test
```

simon-camp (changes requested):
I've created a tracking issue for the disabled test, otherwise LGTM

simon-camp (approved):

NatashaKnk:
This would need to be changed to match up with the TOSA spec, but since we don't have anything running off it yet I think it's okay to submit as-is and I'll send a follow-up PR. Thanks for the heads up!

NatashaKnk (approved):

PR #345, opened 2023-01-27
Integrate LLVM at llvm/llvm-project@435225c

marbre:
Updates LLVM usage to match llvm/llvm-project@435225c. Further updates the MLIR-HLO submodule to tensorflow/mlir-hlo@9fefc2b.


lib/Conversion/MhloToEmitC/MHLORegionOpsToEmitC.cpp
```diff
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BlockAndValueMapping.h"
```
simon-camp, line 17:
```suggestion
```

Can we delete this include?

simon-camp (changes requested):
Just one nit, looks good otherwise.

lib/Conversion/MhloToEmitC/MHLORegionOpsToEmitC.cpp
marbre: We can. Good catch!

simon-camp (approved):

PR #353, opened 2023-04-26
Implement an initial StableHLO to EmitC pass

marbre:
This adds an initianl StableHLO to EmitC pass. The pass is compiled if the option `EMITC_ENABLE_HLO` is set. For now, StableHLO ops converted to C++ rely on the reference implementation for MHLO ops. This will change in the future. When adding more ops, docs will be added and code might be moved from the MHLO conversions to the StableHLO conversions to avoid code duplication. If suitable, the MHLO conversions will be replaced by the StableHLO ones.


simon-camp (approved):
LGTM :+1: 

We can start sharing the conversion patterns when we add support for more StableHLO ops.

marbre:
> We can start sharing the conversion patterns when we add support for more StableHLO ops.

Yes, I am already on it :)

PR #354, opened 2023-04-26
Deduplicate `CallOpConversion`

marbre:
Further renames `CallOpConversion` to `GenericOpConversion`.


simon-camp (commented):
LGTM. Should we rename the conversion pattern to something like GenericOpConversion while we are at it?

marbre:
> LGTM. Should we rename the conversion pattern to something like GenericOpConversion while we are at it?

We agreed to rename it to `GenericOpConversion`.

simon-camp (approved):

PR #356, opened 2023-04-27
Fix include guard

simon-camp:
Co-authored-by: Marius Brehler <marius.brehler@iml.fraunhofer.de>


lib/Conversion/PassDetail.h
```diff
@@ -6,8 +6,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef DIALECT_EMITC_CONVERSION_PASSDETAIL_H
-#define DIALECT_EMITC_CONVERSION_PASSDETAIL_H
+#ifndef CONVERSION_PASSDETAIL_H
+#define CONVERSION_PASSDETAIL_H
```
marbre, line 10:
```suggestion
#ifndef EMITC_CONVERSION_PASSDETAIL_H
#define EMITC_CONVERSION_PASSDETAIL_H
```
Good fetch! However, I assume the suggested include guard might be used in other projects.

Maybe we should make this *more unique*, like Torch-MLIR does, see [here](https://github.com/llvm/torch-mlir/blob/c8e062fb4e8de4620b6ae5e145a9da9a7cca8e5c/lib/Conversion/PassDetail.h#L10-L11). There, `TORCHMLIR_` is used as prefix which closely resembles the project name **and** the name of the include directory (`torch-mlir`). Suggesting `EMITC` to match with our include dir.

lib/Conversion/PassDetail.h
```diff
@@ -23,4 +23,4 @@ class EmitCDialect;
 } // namespace emitc
 } // namespace mlir
 
-#endif // DIALECT_EMITC_CONVERSION_PASSDETAIL_H
+#endif // CONVERSION_PASSDETAIL_H
```
marbre, line 26:
```suggestion
#endif // EMITC_CONVERSION_PASSDETAIL_H
```

marbre (changes requested):

PR #361, opened 2023-05-08
Add .get() method to Tensor<...>.

jacob-hegna:
This is used in the MGLO LLVM integration of emitc'ed models. Our infrastructure internally uses T* buffers to communicate between LLVM and the model (this was used to abstract away the details of TF/TFLite runtime vs interacting with a model in a python host vs EmitC) and so we need to be able to extract the underlying buffer of a EmitC Tensor.

This places a (light) data layout restriction on the Tensor implementation, as we need it to be in a columnar layout. This is already what the implementation did, so there is nothing too surprising here. I have added a unit test to enforce this restriction on the underlying buffer layout.

### DIFF
--- a/reference-implementation/include/emitc/types.h
+++ b/reference-implementation/include/emitc/types.h
@@ -114,7 +114,7 @@ public:
     assert(data.size() == size());
   }
 
-  T* get() { return data.data(); }
+  T *get() { return data.data(); }
 
   static constexpr size_t dim(size_t index) {
     assert(0 <= index && index < rank());
--- a/reference-implementation/include/emitc/types.h
+++ b/reference-implementation/include/emitc/types.h
@@ -114,6 +114,8 @@ public:
     assert(data.size() == size());
   }
 
+  T* get() { return data.data(); }
+
   static constexpr size_t dim(size_t index) {
     assert(0 <= index && index < rank());
     constexpr std::array<size_t, rank()> s = {Shape...};
--- a/reference-implementation/include/emitc/types.h
+++ b/reference-implementation/include/emitc/types.h
@@ -100,6 +100,8 @@ struct switch_t<case_t<B, T>> {
 };
 } // namespace detail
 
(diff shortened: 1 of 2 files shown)


marbre (changes requested):
Thanks for submitting this! I am happy to merge this if it is required for MGLO after the formatting issues are fixed. Furthermore, please add the line `// Copyright Google LLC` to `types.cpp` similar as in `types.h`.

marbre:
Are you normally committing with @google.com or with @gmail.com? If you normally use your work affiliated @google.com address, I would kindly like to ask you to use the one here as well. So far your commit seem to be authored with @gmail.com address.

simon-camp (changes requested):
Can you please add a short description on top of the tensor class? Something like

```
/// The elements of the Tensor are stored contiguously in memory in a row-major layout.
```

marbre:
> Sorry for the delay, I believe all the comments have now been addressed.

No worries :) However, I am currently technically on vacation which will delay my review a little. Sorry for this.

reference-implementation/include/emitc/types.h
```diff
@@ -100,6 +100,8 @@ struct switch_t<case_t<B, T>> {
 };
 } // namespace detail
 
+/// The elements of a Tensor are stored contiguously in memory in a
+/// column-major layout.
```
simon-camp, line 104:
```suggestion
/// row-major layout.
```

The indexing is done using the row-major layout, i.e. the elements of the last dimension are closest together in memory.

jacob-hegna: I see, changed!

simon-camp (approved):

marbre (approved):

marbre:
As you @jacob-hegna sent #366, I am sure you're fine with merging your changes into the re-licensed code base (the original PR applied to code licensed without the LLVM exceptions).

PR #366, opened 2023-05-18
Relicense the reference implementation.

jacob-hegna:
... to Apache2 with LLVM exceptions.

PR #367, opened 2023-05-30
Relicense under Apache-2.0 WITH LLVM Exceptions

marbre:
To facilitate an easy movement between MLIR-EmitC and upstream LLVM and MLIR, respectively, we are relicensing the reference implementation under Apache-2.0 WITH LLVM Exceptions.


marbre:
CC @jacob-hegna and @mtrofin.

In addition to #366, this also relicenses all unit tests.

simon-camp (approved):

PR #368, opened 2023-06-19
Fix `Tensor` with element type bool

henri-gruender:
Introduce classes `TypeWrapper` and `IteratorWrapper` to prevent use of the template specialization for `std::vector<bool>` in the `Tensor` class.

### DIFF
--- a/reference-implementation/include/emitc/types.h
+++ b/reference-implementation/include/emitc/types.h
@@ -91,7 +91,8 @@ struct switch_t<case_t<B, T>> {
 
 // Wrapper class to prevent use of the template specialization for
 // std::vector<bool> in the Tensor class
-template <typename T> class TypeWrapper {
+template <typename T>
+class TypeWrapper {
 public:
   TypeWrapper() = default;
 
@@ -110,7 +111,8 @@ private:
   T value;
 };
 
-template <typename T, typename Iterator> class IteratorWrapper {
+template <typename T, typename Iterator>
+class IteratorWrapper {
 public:
   using iterator_category = std::forward_iterator_tag;
   using value_type = T;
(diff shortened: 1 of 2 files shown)


reference-implementation/include/emitc/types.h
```diff
@@ -88,25 +88,104 @@ struct switch_t<case_t<B, T>> {
...
+  using storage_type = detail::TypeWrapper<T>;
+
 public:
   using value_type = T;
-  using reference = typename std::vector<T>::reference;
-  using iterator = typename std::vector<T>::iterator;
-  using const_iterator = typename std::vector<T>::const_iterator;
+  static_assert(sizeof(T) == sizeof(storage_type),
+                "storage_type has a different size than the raw type T");
+  static_assert(std::is_standard_layout_v<storage_type>,
```
simon-camp, line 173:
```suggestion
  static_assert(std::is_standard_layout<storage_type>::value,
```
I don't think we have the defined the language standard of the generated code anywhere.

marbre: I don't think we've set the C++ standard, but C++14 is reasonable. Depending on the code we generate in the conversion paths it is possible to require C++20 (see https://mlir.llvm.org/docs/Dialects/EmitC/), but without further checking I don't know what our reference implementation requires so far.

henri-gruender: Since this change just resolves syntactic sugar I will still apply it. Also because it might fix the failed workflow run.

marbre:
It seems the stablehlo test is stuck. @henri-gruender have you executed this one locally?

henri-gruender:
I can possibly reproduce the issue and am looking into it. For me it hangs on the [compilation of the StableHLO e2e test](https://github.com/iml130/mlir-emitc/blob/81e189ded5d330393de89a1e240d22e535519e04/.github/workflows/build-and-test.yaml#L226) or takes a very long time.

henri-gruender:
I am unable to resolve the problem in an acceptable way and will close the pull request for the time being.

PR #388, opened 2023-08-30
Refactor tosa unittests

henri-gruender:
Refactor unittests for the following tosa ops:
* arithmetic_right_shift
* equal
* greater_equal
* slice
* reshape
* reduce_sum
* logical_left_shift


reference-implementation/unittests/tosa.cpp
```diff
@@ -296,49 +296,239 @@ TEST(tosa, rescale) {
...
+    Tensor4D<int16_t, 2, 1, 2, 2> expected_result{
+        -0x80, 0, 0, -0x10, -0b1000, 0b1000, -0b110, 0b100};
+    Tensor4D<int16_t, 2, 1, 2, 2> result = tosa::logical_left_shift(x, y);
+
+    EXPECT_THAT(result, Pointwise(Eq(), expected_result));
+  }
 }
 
-TEST(tosa, mul) {
+TEST(tosa, mul) { //....
```
marbre, line 531:
```suggestion
TEST(tosa, mul) {
```
Seems like this line was accidentally changed / added to the commit.

henri-gruender: Correct. Thanks for catching that.

PR #401, opened 2024-01-10
Integrate LLVM at llvm/llvm-project@b2cdf3cc

marbre:
Updates LLVM usage to match llvm/llvm-project@b2cdf3cc. Further updates the StableHLO submodule to openxla/stablehlo@3bdd859.


simon-camp (dismissed):

simon-camp (approved):
LGTM, we should use the rewriter for the attribute conversion in the other places as well. I filed #402  to track this.

PR #412, opened 2024-02-20
Update information on Func to EmitC

marbre:


simon-camp (dismissed):
LGTM. Though I see we mention support for `func.constant` ops. I know there is an overload for this operation in the emitter, but it will always fail as function types cannot be emittted. I guess it was introduced in the std dialect split. We should fix this upstream.

marbre:
Technically `emitc.func` is still *supported* in the pinned version, but I removed it entirely from upstream. Updated the documentation therefore.

simon-camp (approved):

PR #415, opened 2024-04-04
Refactor type of shift argument passed to tosa.rescale

### LINKED ISSUE #406: Refactor type of shift argument passed to tosa.rescale
With #405, the TOSA spec changed the type of the shift argument from `i32` to `i8`. The reference implementation still expects an `int32_t` whereas an` int8_t` is passed in the generated code. Consequently, the type in the reference implementation should be refactored from `int32_t` to `int8_t`.

henri-gruender:
Closes #406


simon-camp (approved):
Thanks, LGTM

marbre (approved):

PR #424, opened 2024-07-23
Adding support for stablehlo::rsqrt

wolfcomos:
Added support for stablehlo::rsqrt, moved reciprocal op to core_ops.h so it can support for stablehlo and tosa.


simon-camp:
I'll close this PR as this project is no longer being actively maintained.

